PR #162, opened 2025-05-13
support jinja2 templates in pipeline definition

xwu2intel:
### Description

This serves as a POC for supporting jinja2 template in the pipeline definition. Simplify pipeline definition and make it align better with the gstreamer command line. See JIRA ITEP-25482 for more details. The Jinja2 support is backward compatible with the existing json schema. 

Jinja2, of the BSD license, is used in the code.

### How Has This Been Tested?

Unit tested:
```
# config.json
{
      "name": "pallet_defect_detection_default",
      "source": "gstreamer",
      "queue_maxsize": 50,
      "pipeline": "filesrc location={{ video_file }} ! decodebin ! videoconvert ! gvadetect name=detection pre-process-backend=opencv model={{ model_name }} model-instance-id=detshared ! queue ! gvawatermark ! gvafpscounter ! gvametaconvert add-empty-results=true name=metaconvert ! gvametapublish name=destination file-path={{ results_json }} file-format=json-lines ! appsink name=appsink wait-on-eos=false",
      "auto_start": false
    }
}
```

```
# curl command line:
curl <URL> -X POST -H 'Content-Type:Application/json' -d '{"video_file":"/home/pipeline-server/resources/videos/warehouse.avi","model_name":"/home/pipeline-server/resources/models/geti/pallet_defect_detection/deployment/Detection/model/model.xml","results_json": "/var/cache/pipeline_root/results-1.jsonl"}'
```

### Checklist:

### DIFF
--- a/microservices/dlstreamer-pipeline-server/src/server/pipeline_manager.py
+++ b/microservices/dlstreamer-pipeline-server/src/server/pipeline_manager.py
@@ -16,6 +16,8 @@ import jsonschema
 from src.server.common.utils import logging
 from src.server.pipeline import Pipeline
 from src.server import schema
+import jinja2
+import copy
 
 class PipelineManager:
 
@@ -107,7 +109,7 @@ class PipelineManager:
                                         config['version'] = version
                                         # validate_config will throw warning of
                                         # missing elements but continue execution
-                                        self._validate_config(config)
+                                        # self._validate_config(config)
                                         self.logger.info("Loading Pipeline: {} version: "
                                                          "{} type: {} from {}".format(
                                                              pipeline,
@@ -272,7 +274,10 @@ class PipelineManager:
             return None, "Invalid Pipeline or Version"
 
         pipeline_type = self.pipelines[name][str(version)]['type']
-        pipeline_config = self.pipelines[name][str(version)]
+        pipeline_config = copy.deepcopy(self.pipelines[name][str(version)])
+
+        if 'template' in pipeline_config:
+            pipeline_config['template'] = jinja2.Environment(undefined=jinja2.StrictUndefined).from_string(pipeline_config['template']).render(request_original)
 
         request = request_original.copy()
 
--- a/microservices/dlstreamer-pipeline-server/src/server/requirements.txt
+++ b/microservices/dlstreamer-pipeline-server/src/server/requirements.txt
@@ -1,2 +1,2 @@
 jsonschema[format_nongpl] == 3.2.0
-jinja2 == 3.1.6
+jinja2 >= 3.1.6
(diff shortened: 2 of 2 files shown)


microservices/dlstreamer-pipeline-server/src/server/pipeline_manager.py
```diff
@@ -107,7 +109,7 @@ def _load_pipelines(self):
                                         config['version'] = version
                                         # validate_config will throw warning of
                                         # missing elements but continue execution
-                                        self._validate_config(config)
+                                        # self._validate_config(config)
```
tjanczak, line 112: I think we should still validate the pipeline; does the validation function not support ninja templates?

ajagadi1 (commented):
Thanks for this PR @xwu2intel! We validated it and it works. However, as @tjanczak mentioned we perhaps still need self._validate_config  - we will check from our side too.

microservices/dlstreamer-pipeline-server/src/server/pipeline_manager.py
xwu2intel: - The _validate_config function does not do much. I suggest to remove the code completely. When used with the dlstreamer syntax, eventually everything is a string. There is no big point of checking for validity.
- In the json schema, when used with property-elements, the format is pretty much free styled. There is not much json validation that can be performed. Of course a developer can define very detailed json validation schema and create a default value and range for every variable. I just wonder how many developers will actually do this as it is very complicated and the pipeline definition is further deviated from the original dlstreamer syntax. 
- At parse_launch, the underlying dlstreamer plugins will perform validity and range checks. The REST API will abort as a result. It's perfectly fine. Thus _validate_config is again redundant. 
- When used with a jinja2 template, we have to delay validation until parse_launch, as the template is expanded at that time. Thus _validate_config does not apply for jinja2 templates at earlier stages.

xwu2intel: @tjanczak

xwu2intel:
> apologies on the late response.
> 
> I think jinja2 template is helpful only when your REST payload adds/updates element properties. such as location, file-path, model, etc. But please know that today the REST payload also supports some data that are not element properties such as custom publishers such as S3 Publisher, OPCUA, ROS2 to be specific pipelines with appsink=destination. See [docs](https://docs.openedgeplatform.intel.com/edge-ai-libraries/dlstreamer-pipeline-server/main/user-guide/advanced-guide/detailed_usage/publisher/Overview.html#publishers) For them we still have to follow the usual way of launching pipelines since they not just carry properties but during launch also make pipeline-server aware of the publisher itself. Also, for RTSP and WebRTC, secondary pipelines I'm not sure how that can be enabled with jinja2 definition.
> 
> Although helpful, having a jinja2 template to support only a subset of pipeline launch behavior is somewhat concerns me.
> 
> cc @tjanczak @rrajore @xwu2intel

The jinja template applies only to the `pipeline` string, which gives the ability to combine CPU/GPU/NPU definitions, a for-loop for multiple streams. The rest definitions, like s3, opcua, ros2, remain as is. They can be configured from separate key/values in json. What's the issue here?

ajagadi1 (approved):

PR #177, opened 2025-05-19
Improve benchmark algo

oommensy:
### Description

Changed Algorithm from (Exponential Doubling + Binary Search):
1. Run 1 stream.
2. estimate = total fps of 1 stream/floor
3. Run "estimate" stream.
4. If per stream fps is more than floor, repeat step (2,3)
5. If per stream fps is less, estimate = estimate -1 and Step 3.

Fixes # (issue)

No

### How Has This Been Tested?

Tested with Floor = 30, 20,50
Tested with Rate = 20%, 0%, 100%

Number of computations reduced from 12 to 4

### Checklist:


tjanczak (commented):
Can we simply use script from: https://github.com/intel-retail/performance-tools/blob/main/benchmark-scripts/stream_density.sh?

We should not have two different methods to measure stream density, please make sure Retail AI Suite and Metro AI Suite (VIPPET) use same logic.

dnoliver:
@tjanczak thanks for the input. 

After reviewing the stream_density.sh implementation, we are going to make this implementation match the one in the shell script. This is the pseudo code we aim to implement:

```
N = 1
set Target FPS

Start Incrementing
Forever
  Start N Pipelines
  Measure Total FPS
  Set Per Stream FPS = Total FPS / N

  If we are Incrementing
    If Per Stream FPS >= Target FPS
        Set Increments = Per Stream FPS / Target FPS

        if Increments <= 1:
            Set Increments = 5
        
    Else
        Start Decrementing
        Set Increments = -1
 
  If we are Decrementing
    If Per Stream FPS >= Target FPS
        Break Success
    Else
        If N <= 1
            Beak Fail

  N = N + Increments
```

dnoliver (changes requested):
Requested to change the algorithm to match the retail benchmarking algorithm

basakcap (approved):

PR #324, opened 2025-06-11
.github: Added persist-credentials flag to time series pull req workflow

vkb1:
### Description

Add the flag to address the low severity issue reported on the workflow for the checkout action

Fixes # (issue) NA

NA

### How Has This Been Tested?

Yes, will wait for the PR workflow to succeed before merging

### Checklist:


.github/workflows/timeseries-build-pull-request.yml
```diff
@@ -47,9 +47,10 @@ jobs:
             docker rm $(docker ps -aq)
         fi
         docker images --quiet --filter=dangling=true | xargs --no-run-if-empty docker rmi -f
-    - uses: actions/checkout@v1
```
sathyendranv, line 50: If Version is changed from v1 then, Please change the path for all other steps below as ./**PATH**/microservices/.. 

Eg. Line55 (https://github.com/open-edge-platform/edge-ai-libraries/pull/324/files#diff-908740f4be8b25801a75f86fd8aaba2897e85e45595f288c761c78dc343b9b7eL55) 
should be changed to cd ./timeseries/microservices/time-series-analytics/docker

To Test the workflow you can introduce a space in some file for our microservice which will trigger the PR check flow

vkb1: Oh, I wasn't aware of this checkout behavior. I've reverted the change now.

sathyendranv (approved):

PR #336, opened 2025-06-12
Remove root user and Add appuser for Docker container and set ownersh…

bhardwaj-nakul:
### Description

Remove root user and Add appuser for Docker container and set ownership for model cache directories.

Fixes # (issue)
CIS docker images scan highlighted use of root user. This PR aims to fix this issue by adding a non-root 'appuser' and granting it relevant ownership of required directories.

### Checklist:

### DIFF
--- a/microservices/vlm-openvino-serving/docker/Dockerfile
+++ b/microservices/vlm-openvino-serving/docker/Dockerfile
@@ -18,6 +18,8 @@ FROM python:3.12.10-slim
 
 ARG INSTALL_DRIVER_VERSION="24.39.31294"
 
+RUN groupadd -g 1001 appuser && useradd -m -s /bin/bash -u 1001 -g 1001 appuser
+
 RUN apt-get update -y && apt-get install -y --no-install-recommends --fix-missing \
     curl 
 
@@ -34,6 +36,11 @@ RUN apt-get clean && rm -rf /tmp/install_gpu_drivers.sh && rm -rf /var/lib/apt/l
 WORKDIR /app
 COPY . .
 RUN chmod +x /app/scripts/compress_model.sh
+
+RUN mkdir -p /root/.cache/huggingface /app/ov-model && \
+    chown -R appuser:appuser /root/.cache/huggingface /app/ov-model
+
+USER appuser
 
 EXPOSE 8000
 


microservices/vlm-openvino-serving/docker/Dockerfile
```diff
@@ -18,6 +18,8 @@ FROM python:3.12.10-slim
 
 ARG INSTALL_DRIVER_VERSION="24.39.31294"
 
+RUN groupadd -g 1001 appuser && useradd -m -s /bin/bash -u 1001 -g 1001 appuser
```
bharagha, line 21: We use egaiuser or something like that in chatqna. Can we stick to same user for consistency?

microservices/vlm-openvino-serving/docker/Dockerfile
```diff
@@ -35,6 +37,11 @@ WORKDIR /app
 COPY . .
 RUN chmod +x /app/scripts/compress_model.sh
 
+RUN mkdir -p /root/.cache/huggingface /app/ov-model && \
+    chown -R appuser:appuser /root/.cache/huggingface /app/ov-model
+
+USER appuser
+
 EXPOSE 8000
```
bharagha, line 45: Just a check: Please check for docker size and bring in the bkm's to reduce it. We can discuss further if required.

microservices/vlm-openvino-serving/docker/Dockerfile
bhardwaj-nakul: We are using 'intelgai' as username. I didn't use this one as we had removed usage of 'intel' from everywhere. 
Do you think it is fine to use 'intelgai' ?

microservices/vlm-openvino-serving/docker/Dockerfile
bhardwaj-nakul: We have used multi-stage build to reduce docker image size. We can add to backlog to find ways to reduce image size further.

microservices/vlm-openvino-serving/docker/Dockerfile
bharagha: Agree. We should not use intel. Remember doing that comment. Go ahead with this ID. We will change in chatqna to appuser

bharagha (approved):

PR #411, opened 2025-06-26
ITEP-70844 bug fix build from source

14pankaj:
Bug fix for chatqna build from source


sample-applications/chat-question-and-answer/chart/subchart/embedding/tei/values.yaml
```diff
@@ -9,7 +9,7 @@ teiEmbeddingService:
   name: tei-embedding-service
   image: 
     repository: ghcr.io/huggingface/text-embeddings-inference
-    tag: cpu-1.5
+    tag: cpu-ipex-latest
```
bharagha, line 12: Is this a specific version or a tag that will keep moving?
Preferable to refer to a specific version.

14pankaj: This is the version recommended for Intel CPU as per huggingface. The other version available is cpu-1.7 which is incremented from cpu-1.5(earlier version used), cpu-1.6. So, the understanding is the cpu-ipex-latest should always use the latest image and we shall not run into same issue again.

bharagha: Referring to : 
image

Do you know if the tag will be moved to another version and what if there is an issue as a consequence? Why not use 1.7.3 tag?

14pankaj: I too checked it, there is no evidence that suggests this would remain the same. Also, looking at this if the same latest tag is used for future images that can have breaking changes can cause unknown failures. So, going with a specific version is a good option. I will verify the application with cpu-1.7 and cpu-1.7.3.

PR #452, opened 2025-07-02
Add HUGGINGFACE_TOKEN support for gated models in VLM OpenVINO Serving

bhardwaj-nakul:
### Description

This change adds support for the `HUGGINGFACE_TOKEN` environment variable in the VLM OpenVINO Serving microservice. It allows users to set their Hugging Face token to access gated models. The token can be set in the Docker Compose file and is also included in the setup script. The change ensures that the token is passed to the OpenVINO Serving container, enabling it to download and use gated models from Hugging Face.

### Checklist:

### DIFF
--- a/microservices/vlm-openvino-serving/docker/Dockerfile
+++ b/microservices/vlm-openvino-serving/docker/Dockerfile
@@ -20,6 +20,9 @@ ARG INSTALL_DRIVER_VERSION="24.39.31294"
 ARG USER_ID=1000
 ARG USER_GROUP_ID=1000
 
+ARG HUGGINGFACE_TOKEN=none
+ENV HUGGINGFACE_TOKEN=${HUGGINGFACE_TOKEN}
+
 RUN groupadd -g ${USER_GROUP_ID} appuser && useradd -m -s /bin/bash -u ${USER_ID} -g ${USER_GROUP_ID} appuser
 
 RUN apt-get update -y && apt-get install -y --no-install-recommends --fix-missing \
@@ -46,4 +49,4 @@ USER appuser
 
 EXPOSE 8000
 
-CMD ["/bin/bash", "-c", "/app/scripts/compress_model.sh ${VLM_MODEL_NAME} ${VLM_COMPRESSION_WEIGHT_FORMAT} && gunicorn -b 0.0.0.0:8000 -k uvicorn.workers.UvicornWorker src.app:app --timeout 10000 -w ${WORKERS:-1} --log-level info --access-logfile - --error-logfile - --log-config /app/logging.conf"]
+CMD ["/bin/bash", "-c", "/app/scripts/compress_model.sh ${VLM_MODEL_NAME} ${VLM_COMPRESSION_WEIGHT_FORMAT} ${HUGGINGFACE_TOKEN} && gunicorn -b 0.0.0.0:8000 -k uvicorn.workers.UvicornWorker src.app:app --timeout 10000 -w ${WORKERS:-1} --log-level info --access-logfile - --error-logfile - --log-config /app/logging.conf"]
--- a/microservices/vlm-openvino-serving/docker/Dockerfile
+++ b/microservices/vlm-openvino-serving/docker/Dockerfile
@@ -21,7 +21,6 @@ ARG USER_ID=1000
 ARG USER_GROUP_ID=1000
 
 ARG HUGGINGFACE_TOKEN=none
-ENV HUGGINGFACE_TOKEN=${HUGGINGFACE_TOKEN}
 
 RUN groupadd -g ${USER_GROUP_ID} appuser && useradd -m -s /bin/bash -u ${USER_ID} -g ${USER_GROUP_ID} appuser
 
--- a/microservices/vlm-openvino-serving/scripts/compress_model.sh
+++ b/microservices/vlm-openvino-serving/scripts/compress_model.sh
@@ -16,7 +16,6 @@ echo "Model Directory: $MODEL_DIR"
 
 # Login to Hugging Face if token is provided and not 'none'
 if [ -n "$HUGGINGFACE_TOKEN" ] && [ "$HUGGINGFACE_TOKEN" != "none" ]; then
-    # pip install --no-cache-dir huggingface-hub
     echo "Logging in to Hugging Face to access gated models..."
     huggingface-cli login --token "$HUGGINGFACE_TOKEN"
 fi
--- a/microservices/vlm-openvino-serving/scripts/compress_model.sh
+++ b/microservices/vlm-openvino-serving/scripts/compress_model.sh
@@ -3,9 +3,9 @@
 # Copyright (C) 2025 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
-
 VLM_MODEL_NAME=$1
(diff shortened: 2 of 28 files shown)


microservices/vlm-openvino-serving/scripts/compress_model.sh
```diff
@@ -14,6 +14,13 @@ echo "Model Name: $VLM_MODEL_NAME"
 echo "Compression Weight Format: $VLM_COMPRESSION_WEIGHT_FORMAT"
 echo "Model Directory: $MODEL_DIR"
 
+# Login to Hugging Face if token is provided and not 'none'
+if [ -n "$HUGGINGFACE_TOKEN" ] && [ "$HUGGINGFACE_TOKEN" != "none" ]; then
+    # pip install --no-cache-dir huggingface-hub
+    echo "Logging in to Hugging Face to access gated models..."
+    huggingface-cli login --token "$HUGGINGFACE_TOKEN"
```
yogeshmpandey, line 20: If huggingface-cli does not exist on the system?

microservices/vlm-openvino-serving/docker/Dockerfile
```diff
@@ -20,6 +20,9 @@ ARG INSTALL_DRIVER_VERSION="24.39.31294"
 ARG USER_ID=1000
 ARG USER_GROUP_ID=1000
 
+ARG HUGGINGFACE_TOKEN=none
+ENV HUGGINGFACE_TOKEN=${HUGGINGFACE_TOKEN}
```
yogeshmpandey: Why is the build argument needed ? Can we not pass it at runtime ?

microservices/vlm-openvino-serving/setup.sh
```diff
@@ -45,11 +45,22 @@ export VLM_SEED=42
...
 # Check if VLM_MODEL_NAME is not defined or empty
 if [ -z "$VLM_MODEL_NAME" ]; then
     echo -e "ERROR: VLM_MODEL_NAME is not set in your shell environment."
     return
 else
     export VLM_MODEL_NAME=$VLM_MODEL_NAME
     echo -n VLM_MODEL_NAME: ${VLM_MODEL_NAME}
+fi
+if [ -z "$HUGGINGFACE_TOKEN" ]; then
+    echo -e "\nWARNING: HUGGINGFACE_TOKEN is not set."
```
yogeshmpandey: Not able to understand this. This will give a warning every time since we are unsettling the env. Also for the non-Gated models, this will show a warning even though the token is not needed.

sample-applications/video-search-and-summarization/setup.sh
```diff
@@ -98,6 +98,7 @@ export USER_ID=$(id -u)
 export USER_GROUP_ID=$(id -g)
 export VIDEO_GROUP_ID=$(getent group video | awk -F: '{printf "%s\n", $3}')
 export RENDER_GROUP_ID=$(getent group render | awk -F: '{printf "%s\n", $3}')
+export HUGGINGFACE_TOKEN=${HUGGINGFACE_TOKEN}
```
yogeshmpandey: Why is this line needed ? it would already be exported by the user

yogeshmpandey (commented):
Is this change tested on helm deployment as well ?

microservices/vlm-openvino-serving/scripts/compress_model.sh
bhardwaj-nakul: This is inside the container. Running the model conversion script before running fastapi server

microservices/vlm-openvino-serving/docker/Dockerfile
bhardwaj-nakul: I have kept ENV as well so that in future if this token is needed in container env, it is still available

microservices/vlm-openvino-serving/setup.sh
bhardwaj-nakul: This warning is there in the standalone microservice script, not the VSS setup.sh. This is just to indicate to the user whether the huggingface token has been set or not when running microservice standalone.

microservices/vlm-openvino-serving/docker/Dockerfile
yogeshmpandey: I don't think it is a good idea to persist tokens in an image.

microservices/vlm-openvino-serving/scripts/compress_model.sh
yogeshmpandey: ok, meaning that the huggingface-cli exists in the container? If yes ,then we can remove the commented part ?

sample-applications/video-search-and-summarization/setup.sh
bhardwaj-nakul: Followed the same pattern as $VLM_MODEL_NAME. can be avoided as well. This is to just trace of this variable through setup.sh - that this variable is needed by vlm docker compose.

microservices/vlm-openvino-serving/setup.sh
yogeshmpandey: Ideally we would want to show the warning if its a Gated model and `HUGGINGFACE_TOKEN` is not set.

sample-applications/video-search-and-summarization/setup.sh
yogeshmpandey: Let's remove it. we will remove other redundent values in subsequent PRs

microservices/vlm-openvino-serving/scripts/compress_model.sh
bhardwaj-nakul: yes, i had already pushed the commit to remove commented pip install.

microservices/vlm-openvino-serving/setup.sh
bhardwaj-nakul: 2 reasons:

1. we are not pulling model before starting container, which would enable us to prompt the user to provide token if gated model. This is happening in container at runtime, we are expecting the user to pass the token if they are trying to use gated model.

2. Since we are not explicitly pulling the mode here, instead we are passing model name to optimum which internally pulls the model, it is difficult to parse the output and see what error is there if it is not able to pull the model.

microservices/vlm-openvino-serving/docker/Dockerfile
bhardwaj-nakul: it will be passed only through runtime arg, not as build arg. so it is not baked into the image. 

But yes, we can remove this from here, Since it is equivalent to having env params in docker compose

sample-applications/video-search-and-summarization/setup.sh
bhardwaj-nakul: ok

microservices/vlm-openvino-serving/docker/Dockerfile
bhardwaj-nakul: removed from dockerfile, as this param is already defined as env in docker compose

microservices/vlm-openvino-serving/setup.sh
bhardwaj-nakul: Added detailed echo statements explaining huggingface tokens and how to set it

yogeshmpandey (approved):

bharagha (approved):

PR #469, opened 2025-07-04
Document Ingestion env updates & helm changes reverted for permission…

14pankaj:
Updated Document-ingestion ENV to remove intelgai
Update helm chart for Chatqna-ui for Permission issue

### DIFF
--- a/microservices/document-ingestion/pgvector/app/.env
+++ b/microservices/document-ingestion/pgvector/app/.env
@@ -1,5 +1,5 @@
-LOCAL_STORE_PREFIX = "/tmp/intelgai"
-DEFAULT_BUCKET = "intel.gai.ragfiles"
-OBJECT_PREFIX = "intelgai"
+LOCAL_STORE_PREFIX = "/tmp/genai"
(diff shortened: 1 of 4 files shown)


microservices/document-ingestion/pgvector/app/.env
```diff
@@ -1,5 +1,5 @@
-LOCAL_STORE_PREFIX = "/tmp/intelgai"
-DEFAULT_BUCKET = "intel.gai.ragfiles"
-OBJECT_PREFIX = "intelgai"
+LOCAL_STORE_PREFIX = "/tmp/genai"
```
bharagha: why not just user the USER name as prefix. I think it is **appuser**

14pankaj: this is for the file name prefix. Yes we can have appuser as well. Just on the UI the filename will be displayed as appuser_filename_asdfsdf.pdf

bharagha: Point was not to use genai or Intel in name. appuser is ok

PR #479, opened 2025-07-07
fix the service name

14pankaj:
OVMS GPU service name fix in docker compose


bharagha (approved):

bharagha (commented):
Please discuss with @hteeyeoh on the original change reasons...

madhuri-rai07 (approved):

PR #553, opened 2025-07-18
ChatQnA fixes

hteeyeoh:
### Description

This PR fixes several issues:

- ChatQnA and Core UI container run in root privileged
- Nginx container run in root privileged
- ChatQnA Core model configuration loading using YAML

Fixes # (issue)

### How Has This Been Tested?

Validated locally for both docker compose and helm deployment

### Checklist:


sample-applications/chat-question-and-answer-core/app/chain.py
```diff
@@ -14,73 +14,55 @@
...
+# This flag is set in the conftest.py file before running the tests.
+if os.getenv("RUN_TEST") != "True":
     # login huggingface
     login_to_huggingface(config.HF_ACCESS_TOKEN)
 
     # Download convert the model to openvino optimized
-    download_huggingface_model(config.EMBEDDING_MODEL_ID, config.CACHE_DIR)
-    download_huggingface_model(config.RERANKER_MODEL_ID, config.CACHE_DIR)
-    download_huggingface_model(config.LLM_MODEL_ID, config.CACHE_DIR)
+    download_huggingface_model(config.EMBEDDING_MODEL_ID, config._CACHE_DIR)
```
bharagha, line 27: Just to double confirm: We are using _CACHE_DIR instead of CACHE_DIR because it is private? Convention wise _ indicates private right.

sample-applications/chat-question-and-answer-core/app/chain.py
```diff
@@ -14,73 +14,55 @@
 import os
 import pandas as pd
 
-config = Settings()
 vectorstore = None
```
bharagha, line 17: We are now using config as an object. Previously, instantiated per run by using Settings. In a concurrent usage (which ChatQnA Core) is not intended for, there could be potential issues with respect to multiple threads using the same object. Has this been thought through for its implications/impact.

sample-applications/chat-question-and-answer-core/app/chain.py
```diff
@@ -14,73 +14,55 @@
...
 # The RUN_TEST flag is used to bypass the model download and conversion steps during pytest unit testing.
-# By default, the flag is set to 'false', enabling the model download and conversion process in a normal run.
-# To skip these steps, set the flag to 'true'.
-# Check environment flag
-RUN_TEST = os.getenv('RUN_TEST', False)
-
-if not RUN_TEST:
+# If RUN_TEST is set to "True", the model download and conversion steps are skipped.
+# This flag is set in the conftest.py file before running the tests.
+if os.getenv("RUN_TEST") != "True":
```
bharagha: Suggest put a check for lower case too. (Using .lower() kind of thing)

sample-applications/chat-question-and-answer-core/app/chain.py
```diff
@@ -14,73 +14,55 @@
...
-    ---------
-    {context}
 
-    ---------
-    Question: {question}
-    ---------
-
-    Answer:
-    """
+    template = config.PROMPT_TEMPLATE
```
bharagha, line 37: Error check on prompt template? That it is conforming to certain expectations etc. Not critical but just thinking through on the implications.

sample-applications/chat-question-and-answer-core/app/chain.py
```diff
@@ -14,73 +14,55 @@
...
     # login huggingface
     login_to_huggingface(config.HF_ACCESS_TOKEN)
 
     # Download convert the model to openvino optimized
-    download_huggingface_model(config.EMBEDDING_MODEL_ID, config.CACHE_DIR)
-    download_huggingface_model(config.RERANKER_MODEL_ID, config.CACHE_DIR)
-    download_huggingface_model(config.LLM_MODEL_ID, config.CACHE_DIR)
+    download_huggingface_model(config.EMBEDDING_MODEL_ID, config._CACHE_DIR)
+    download_huggingface_model(config.RERANKER_MODEL_ID, config._CACHE_DIR)
+    download_huggingface_model(config.LLM_MODEL_ID, config._CACHE_DIR)
```
bharagha, line 29: No error check required in any of these operations?
Also, error check for config parameters in general is not there.

sample-applications/chat-question-and-answer-core/app/config.py
```diff
@@ -1,50 +1,98 @@
...
+    LLM_MODEL_ID: str = ""
+    PROMPT_TEMPLATE: str = ""
+    EMBEDDING_DEVICE: str = "CPU"
+    RERANKER_DEVICE: str = "CPU"
+    LLM_DEVICE: str = "CPU"
+    MAX_TOKENS: int = 1024
+    ENABLE_RERANK: bool = True
+
+    # These fields will not be affected by environment variables
+    _CACHE_DIR: str = PrivateAttr("/tmp/model_cache")
```
bharagha, line 58: Hardcoded values. Did you consider deployment with PVC or with model download service?

Also, did not get the idea behind using PrivateAttr for these variables though they can be set externally. Just a check from usual conventions perspective.

sample-applications/chat-question-and-answer/ui/react/Dockerfile
```diff
@@ -1,29 +1,54 @@
...
+# Define build-time arguments for user and group IDs.
+# Default to 101, which matches the UID/GID of the 'nginx' user in the final stage.
+# This ensures file ownership compatibility across stages without needing chown.
+# -----------------------------------------------------------------------------
+ARG UID=101
+ARG GID=101
 
-USER user
+# Create a non-root user with the specified UID and GID
+RUN groupadd -g ${GID} user && \
```
yogeshmpandey, line 16: Why are we adding this user in build layer. While the non-root user is nginx ?

yogeshmpandey (changes requested):

sample-applications/chat-question-and-answer-core/app/config.py
```diff
@@ -1,50 +1,98 @@
...
+    _HF_DATASETS_CACHE: str = PrivateAttr("/tmp/model_cache")
+    _TMP_FILE_PATH: str = PrivateAttr("/tmp/chatqna/documents")
+    _DEFAULT_MODEL_CONFIG: str = PrivateAttr("/tmp/model_config/default_model.yaml")
+    _MODEL_CONFIG_PATH: str = PrivateAttr("/tmp/model_config/config.yaml")
+
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+        # The RUN_TEST flag is used to bypass the model config loading during pytest unit testing.
```
bharagha, line 68: We need to document the precedence between env, yaml, and constructor arguments as in who can overwrite whom...

sample-applications/chat-question-and-answer-core/app/config.py
```diff
@@ -1,50 +1,98 @@
...
+
+        for section in ("model_settings", "device_settings"):
+            for key, value in config.get(section, {}).items():
+                if hasattr(self, key):
+                    setattr(self, key, value)
+
+        if not self.PROMPT_TEMPLATE:
+            print("INFO - PROMPT_TEMPLATE is not set. Get prompt template based on LLM_MODEL_ID.")
+            self.PROMPT_TEMPLATE = get_prompt_template(self.LLM_MODEL_ID)
+
```
bharagha, line 106: What if LLM_MODEL_ID is not set before this which could happen in concurrent runs?

sample-applications/chat-question-and-answer-core/app/document.py
```diff
@@ -9,8 +9,6 @@
     TextLoader
```
bharagha, line 9: In general, I guess assumption is that this application is to be used in single user/non-concurrent mode only. We should document the same.

sample-applications/chat-question-and-answer/ui/react/Dockerfile
hteeyeoh: this is added so that the it having the same UID and GID (101) for the output for UI compilation and this will make sure in the 2nd stage when these output and copy to nginx non-root image it wont bump into permission issue.

sample-applications/chat-question-and-answer-core/model_config/intel-neural-chat-7b-v3-3-config.yaml
```diff
@@ -0,0 +1,21 @@
+model_settings:
```
bharagha, line 1: Should we check if any AI assessment is required now that we have model specific suggested configuration?

sample-applications/chat-question-and-answer-core/app/chain.py
hteeyeoh: Yes. Pydantic requires private attributes to start with an underscore (_). If you define a PrivateAttr without the underscore, it will not be treated as private, and Pydantic may raise an error or behave unexpectedly

sample-applications/chat-question-and-answer-core/app/config.py
hteeyeoh: I missed out that. Let me check.

sample-applications/chat-question-and-answer-core/app/config.py
hteeyeoh: this is the path for internal container itself not the host itself. So this wouldnt affect if user want to mount the volume in their host system with other paths. Purposely set it to private and hardcoded it in the container just as a constant value to store the model cache inside the container path.

sample-applications/chat-question-and-answer-core/app/chain.py
hteeyeoh: for this config._CACHE_DIR this value is set with a default constant value to store the model in the defined path in container. The underlying function download_huggingface_model and openvino will throw error if any issue happens

sample-applications/chat-question-and-answer-core/app/chain.py
hteeyeoh: correct. I missed that out. Will check for required placeholders in the template to ensure it meets the RAG prompt template criteria

sample-applications/chat-question-and-answer-core/app/chain.py
hteeyeoh: Sure will include this.

sample-applications/chat-question-and-answer-core/app/document.py
hteeyeoh: i thought all the while is single user/non-concurrent mode right?

sample-applications/chat-question-and-answer-core/model_config/intel-neural-chat-7b-v3-3-config.yaml
hteeyeoh: Not really understand the comment here.

sample-applications/chat-question-and-answer-core/app/document.py
bharagha: It is. We have not captured the same in documentation... Maybe in get-started or overview or as a limitation in release note...

sample-applications/chat-question-and-answer-core/app/chain.py
bharagha: Ensure we are treating the private attributes as such in the implementation throughout...

sample-applications/chat-question-and-answer-core/app/chain.py
bharagha: Let's update the document to say concurrent usage is not supported and close this comment.

hteeyeoh: The major reason I make this update is considering the efficiency. This avoids repeated parsing/loading across application. 

Currently the Settings class itself is mostly immutable. Since the current chatqna-core implementation does not let user to modify the Settings class object after initialization, so it's safe to share across threads. This settings object are kinda "read-only" object after initialization. User are not provided with any endpoint of function to modify the config parameter unless they restart the application and update the yaml.

So, from my POV it is at least safe for now going with this approach unless there's some design and implementation changes for chatqna-core in the future.

sample-applications/chat-question-and-answer-core/model_config/intel-neural-chat-7b-v3-3-config.yaml
bharagha: Let's talk

sample-applications/chat-question-and-answer-core/app/document.py
hteeyeoh: Maybe you can suggest the line for me I can include in this PR in get-started doc.

sample-applications/chat-question-and-answer-core/app/chain.py
hteeyeoh: > Let's update the document to say concurrent usage is not supported and close this comment.

If you prefer the original implementation, instantiated per run by using Settings it also fine. At least i dont think now will have any difference.

sample-applications/chat-question-and-answer-core/app/chain.py
hteeyeoh: Definitely. This is rules of Pydantic if we are using PrivateAttr

sample-applications/chat-question-and-answer-core/app/chain.py
hteeyeoh: Done.

sample-applications/chat-question-and-answer-core/app/config.py
hteeyeoh: Done adding checks

sample-applications/chat-question-and-answer-core/app/chain.py
hteeyeoh: Added the prompt_template validation to check for required placeholders such as {question} and {context} to meet requirements for RAG prompt template

sample-applications/chat-question-and-answer-core/app/document.py
hteeyeoh: Updated the get started doc as per your suggestion offline.

sample-applications/chat-question-and-answer-core/app/config.py
hteeyeoh: We can take this as improvement in future if any need. As for now, those values required from user already documented in the guide. The rest which is constant/not exposed to user already set as private_attribute in the class via fastapi private attr feat,

madhuri-rai07 (approved):

PR #627, opened 2025-08-01
Nginx UI bandit fix

Supriya-Krishnamurthi:
This PR includes changes from 3 PRs

-> Bandit fixes

-> Disable Generate Summary button and upload file option when a summary is being generated such that it doesnt allow users to upload any other file during that phase or re-trigger summary generation

-> Nginx Root user issue is fixed
    
    nginx.conf:

    -> listen 8080; - nginx listening on port 8080 internally (non-root safe, bcz port <1024 are privileged ports)
    -> pid /tmp/nginx.pid; - non-root compatible PID file location

    docker-compose.yaml:

    -> image: nginxinc/nginx-unprivileged:1.27.4 - non-root nginx image
    -> ports: "8101:8080" - mapping external port 8101 to internal port 8080

     Ran CIS report, not seeing "running as root for nginx" error with this change, and app also works fine (along with OTEL)


sample-applications/document-summarization/ui/gradio_app.py
```diff
@@ -60,8 +59,8 @@ def summarize_document(file_obj, custom_query=None):
         
         logger.info(f"Sending request to {docsum_endpoint} with query: {query}")
         
-        # Send the request to the API
-        response = requests.post(docsum_endpoint, files=files, data=data)
+        # Send the request to the API with timeout
+        response = requests.post(docsum_endpoint, files=files, data=data, timeout=300)
```
madhuri-rai07: 300 sec timeout is not enough IMO. It should be atleast -15-20 mins or higher.
@bharagha Can you provide your input?

hteeyeoh (commented):
LGTM. Will approve after the comment is resolved

sample-applications/document-summarization/ui/gradio_app.py
bharagha: Do we give a feedback to user that processing is in progress? Timeout should be worse case normally if this continuous progress update is provided. So timeout value of 900 or 1200 is also ok.

14pankaj: Lets make it configurable as ENV and which user can update.  We can only default that env to a certain value

bharagha: Continuous update on progress is required. Timeout should be worsecase

hteeyeoh (approved):

sample-applications/document-summarization/ui/gradio_app.py
madhuri-rai07: removed the timeout for now. We will add it as improvement for 1.3 release

sample-applications/document-summarization/app/server.py
```diff
@@ -24,6 +25,8 @@
 from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
 from opentelemetry.instrumentation.fastapi import FastAPIInstrumentor
 
+tmp_docs_dir = os.path.join(tempfile.gettempdir(), "docs")
```
bharagha, line 28: Need to check this on EMT in terms of if the app can get the default temp dir. Same should work in docker/helm environments.

sample-applications/document-summarization/docker-compose.yaml
```diff
@@ -1,12 +1,12 @@
 services:
   nginx:
-    image: nginx:latest
+    image: nginxinc/nginx-unprivileged:1.27.4
```
bharagha, line 3: BOM updated right?

madhuri-rai07: New image used here is approved. Will update the BOM

sample-applications/document-summarization/app/server.py
madhuri-rai07: EMT is verified, both docker-compose and helm

bharagha (approved):

PR #687, opened 2025-08-08
Docsum - Updated FAQ doc for OVMS QnA

Supriya-Krishnamurthi:
Docsum - Updated FAQ doc for OVMS QnA


bharagha (approved):

hteeyeoh:
> > Just a minor doubt. Shall we include this following:
> > ### Q: What is the ideal time for services or pods to become ready when deployed via Helm?
> > A: The typical initialization time for services and pods deployed using Helm is approximately 6 to 8 minutes, depending on the system resources and network conditions.
> > This might have chance of misleading user that pods will be up within the time mentioned but this may be vary or depends to user system and network bandwidth right?
> 
> @hteeyeoh the intent of the statement you mentioned looks similar to that of the existing statement, can you provide more clarification here.

Ok. Maybe im taking it wrongly. It should be fine i think as long as we mentioned "time taken depending on the system resources and network conditions." .

hteeyeoh (approved):

PR #771, opened 2025-08-14
Add nltp package

madhuri-rai07:
### Description

Fixes # (issue)

### How Has This Been Tested?

### Checklist:


sample-applications/document-summarization/Dockerfile
```diff
@@ -22,13 +22,19 @@ RUN poetry config virtualenvs.create false \
     && poetry install --only=main --no-root \
     && poetry cache clear pypi --all
 
+# install nltk data
+RUN pip install nltk \
```
hteeyeoh: is there specific reason to handle this nltk separately using pip? cant we include in pyproject and handle it with poetry?

14pankaj (approved):

sample-applications/document-summarization/Dockerfile
```diff
@@ -22,13 +22,19 @@ RUN poetry config virtualenvs.create false \
...
+ && mkdir -p /opt/share/nltk_data \
+ && python -m nltk.downloader -d /opt/share/nltk_data stopwords punkt
```
hteeyeoh, line 27: how is this difference with the line in base.py below:

nltk.download('stopwords')

both are downloading the stopward packages right?

sample-applications/document-summarization/Dockerfile
madhuri-rai07: Moved it to pyproject.toml

sample-applications/document-summarization/Dockerfile
madhuri-rai07: this is to download the files as part of docker image itself, whereas the code in base.py is for development purpose if not using docker.

hteeyeoh: okay. which mean it would be redundant right if using in docker way as same package will download twice to different directory. If this is the case isnt that just keeping it in base.py would be enough?

madhuri-rai07: Just having it in dockerfile is enough, prebaked in image is better.

hteeyeoh: agree with that.

hteeyeoh (approved):

PR #867, opened 2025-08-28
[DLStreamer] Enable gvagenai binary in DLStreamer releases of Debian package and Docker image for MiniCPMv2.6

marcin-wadolkowski:
### Description

Fixes # (issue)

### How Has This Been Tested?

### Checklist:


tbujewsk:
1. What about support for Fedora/Windows ? Pls check whether it is sth that we can easily add .
2. Missed support for the latest way of compilation i.e. 'make build'
3. Update documentation if necessary

marcin-wadolkowski:
> 1. What about support for Fedora/Windows ? Pls check whether it is sth that we can easily add .
> 2. Missed support for the latest way of compilation i.e. 'make build'
> 3. Update documentation if necessary

1. OpenVINO GenAI is only for Ubuntu
2. I don't understand
3. OK

yunowo:
OpenVINO GenAI is only installed in `dlstreamer-dev` image but not in `dlstreamer` image ?

libraries/dl-streamer/Makefile
```diff
@@ -48,7 +51,21 @@ build: dependencies ## Compile Deep Learning Streamer
 		-DENABLE_RDKAFKA_INSTALLATION=ON \
 		-DENABLE_VAAPI=ON \
 		-DENABLE_SAMPLES=ON \
-		-DENABLE_TESTS=OFF 
+		-DENABLE_GENAI=ON \
+		-DENABLE_TESTS=OFF
+	else
```
tjanczak: why do we need if/else?
would 'else' path also work on Ubuntu?

libraries/dl-streamer/docker/ubuntu/ubuntu24.Dockerfile
```diff
@@ -297,6 +297,12 @@ RUN \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
+# OpenVINO Gen AI
+ARG OPENVINO_GENAI_VER=openvino_genai_ubuntu24_2025.2.0.0_x86_64
```
tjanczak: Can we derive {OPENVINO_GENAI} from {OPENVINO_BASE}? 
I think we want to use base and genai which are in sync.

libraries/dl-streamer/docs/source/get_started/install/install_guide_ubuntu.md
```diff
@@ -244,6 +244,27 @@ C. List available versions
 sudo apt show -a intel-dlstreamer
 ```
 
+### [Optional] Step 7: OpenVINO Gen AI installation
+
+To use [gvagenai element](https://docs.openedgeplatform.intel.com/oep/edge-ai-libraries/dl-streamer/elements/gvagenai.html) there is need to:
+
+A. Follow instructions to download [OpenVINO Gen AI archive](https://docs.openvino.ai/2025/get-started/install-openvino/install-openvino-genai.html)
```
tjanczak: Please insert exact commands a user can 'copy-paste'... it should be a single-click installation.
Right now this step requires browsing a few webpages + copy paste of several commands.

libraries/dl-streamer/src/monolithic/gst/elements/CMakeLists.txt
```diff
@@ -101,6 +101,6 @@ add_subdirectory(gvawatermark3d)
 if(ENABLE_REALSENSE)
     add_subdirectory(gvarealsense)
 endif()
-if(ENABLE_GENAI)
+if(ENABLE_GENAI OR WIN32)
```
tjanczak: Please remove WIN32, genai should be installed based on "ENABLE_GENAI" only.

libraries/dl-streamer/Makefile
marcin-wadolkowski: we need to check if OS is Ubuntu. On Fedora we cannot build with GenAI plugin because there is no such OpenVINO GenAi version

libraries/dl-streamer/docker/ubuntu/ubuntu24.Dockerfile
marcin-wadolkowski: I didn't derived because there is one more 0 in version: 2025.2.0.0 but OpenVINO is 2025.2.0 which may be confusing.

libraries/dl-streamer/docs/source/get_started/install/install_guide_ubuntu.md
marcin-wadolkowski: I don't think it's a good idea. If OpenVINO GenAI documentation changes we need to do the same update on our side.

libraries/dl-streamer/src/monolithic/gst/elements/CMakeLists.txt
marcin-wadolkowski: I cannot find easy way to pass ENABLE_GENAI=ON to cmake argument list on Windows.

libraries/dl-streamer/Makefile
tjanczak: I understand now; I'd still recommend to extract difference between Ubuntu and Fedora to a separate if statement - like
if (Ubuntu)
    ENABLE_GENAI=ON
 else
    ENABLE_GENAI=OFF
    
 ... and then use a common cmake with a parameter
    -DENABLE_GENAI={ENABLE_GENAI}

libraries/dl-streamer/docs/source/get_started/install/install_guide_ubuntu.md
tjanczak: We already hard-code these steps inside our docker files... but talking to TomekB - entire section may not be needed. 
We should have genai components installed as part of dlstremer package.

This instruction would still be needed in "Compile from Source" - but not here.

libraries/dl-streamer/src/monolithic/gst/elements/CMakeLists.txt
tjanczak: Let's check with copilot if there is such a way. I think this is global cmake parameter, so we should be able to set default value for Windows.

libraries/dl-streamer/Makefile
```diff
@@ -18,6 +18,14 @@ BUILD_TYPE 			?= Release
 
 DOCKER_PRIVATE_REGISTRY := # Empty on purpose
 
+LINUX_DISTRIBUTION := $(shell lsb_release -ds | cut -d " " -f1)
+ifeq ($(LINUX_DISTRIBUTION), Ubuntu)
+	. /opt/openvino_genai/setupvars.sh
```
pbartosik: Running `setupvars.sh` here doesn't do anything. Makefile creates a fresh shell session for every command and only variables from your shell when calling `make` as well as variables marked with `export` are present in those sub-shells.

mholowni: fixed in 0f736b73c42bf06672ca2d33ab3c44d3b29073c2

mholowni: 2a5fa964d07a3ae8280819ced59f00fe3ada13d2

pbartosik (approved):

libraries/dl-streamer/Makefile
```diff
@@ -11,13 +11,25 @@ nproc ?= 1
...
 
 DOCKER_PRIVATE_REGISTRY := # Empty on purpose
 
+LINUX_DISTRIBUTION := $(shell lsb_release -ds | cut -d " " -f1)
+GENAI_DIR_SET := $(shell if [ -n "$$OpenVINOGenAI_DIR" ]; then echo true; else echo false; fi)
+ifeq ($(LINUX_DISTRIBUTION), Ubuntu)
+ifeq ($(GENAI_DIR_SET), true)
+	ENABLE_GENAI := ON
+else
+	ENABLE_GENAI := OFF
```
tbujewsk: Can't you reduce it ? Move ENABLE_GENAI := OFF out of IF statement and leave only part with ENABLE_GENAI := ON

tbujewsk (approved):

PR #878, opened 2025-08-29
Docsum - Chunked txt file before passing it to model

Supriya-Krishnamurthi:
when we load file with SimpleDirectoryReader() , LlamaIndex uses a PDF reader that:
 
- extract text page by page
- splits data into chunks internally and gives to LLM

For plain text, LlamaIndex just reads the entire file as one document.
For some models like microsoft phi 3.5 mini instruct, inference drags on → request hits your timeout

This PR contains solution to resolve this issue:

- if file extention ends with: TXT → chunk by lines (chunk_text_file).
- For each chunk, SimpleSummaryPack.run(doc_id) → produces a chunk summary.
- Run one more LLM call on the concatenated chunk summaries to distill a final, concise summary.


sample-applications/document-summarization/app/server.py
```diff
@@ -168,30 +185,36 @@ async def stream_data_endpoint(
...
-        
+
+        # 🔹 Load documents (with TXT chunking support)
         try:
-            logger.info("Loading documents from directory")
-            documents = SimpleDirectoryReader(input_files=[file_location]).load_data()
-            documents[0].doc_id = file.filename
-            logger.info(f"Successfully loaded {file_location} document(s)")
+            logger.info("Loading documents")
+            if file.filename.endswith(".txt"):
```
hteeyeoh, line 200: chunking only applied to txt file? how about other file type?

madhuri-rai07 (approved):

sample-applications/document-summarization/app/server.py
madhuri-rai07: other files types are chunked by llamaIndex

hteeyeoh: ok which mean txt file have to handle/chunk explicitly. Got it.

PR #938, opened 2025-09-12
[DLStreamer] gvagenai: enable memory mapper, add support for prompt-path

yunowo:
### Description

Enabled memory mapper for gvagenai so it supports VAMemory, VASurface, DMABuf frames now.
Add support for prompt-path.

None

### How Has This Been Tested?

### Checklist:

### DIFF
--- a/libraries/dl-streamer/docker/fedora41/fedora41.Dockerfile
+++ b/libraries/dl-streamer/docker/fedora41/fedora41.Dockerfile
@@ -67,7 +67,7 @@ RUN \
     gobject-introspection-devel x265-devel x264-devel libde265-devel libgudev-devel libusb1 libusb1-devel nasm python3-virtualenv \
     cairo-devel cairo-gobject-devel libXt-devel mesa-libGLES-devel wayland-protocols-devel libcurl-devel which \
     libssh2-devel cmake git valgrind numactl libvpx-devel opus-devel libsrtp-devel libXv-devel paho-c-devel \
-    kernel-headers pmix pmix-devel hwloc hwloc-libs hwloc-devel libxcb-devel libX11-devel libatomic intel-media-driver libsoup-3.0-0 && \
+    kernel-headers pmix pmix-devel hwloc hwloc-libs hwloc-devel libxcb-devel libX11-devel libatomic intel-media-driver libsoup3 && \
     dnf clean all
 
 RUN \
--- a/libraries/dl-streamer/docker/fedora41/intel-dlstreamer.spec
+++ b/libraries/dl-streamer/docker/fedora41/intel-dlstreamer.spec
@@ -43,7 +43,7 @@ Requires: x265-libs
 Requires: x264-libs
 Requires: openexr
 Requires: tbb
-Requires: libsoup-3.0-0
+Requires: libsoup3
 Requires: intel-media-driver
 Requires: openvino-2025.3.0
 


libraries/dl-streamer/docs/source/elements/gvagenai.md
```diff
@@ -96,6 +96,21 @@ Pad Templates:
                   width: [ 1, 2147483647 ]
                  height: [ 1, 2147483647 ]
               framerate: [ 0/1, 2147483647/1 ]
+      video/x-raw(memory:DMABuf)
+                 format: { (string)DMA_DRM }
+                  width: [ 1, 2147483647 ]
+                 height: [ 1, 2147483647 ]
+              framerate: [ 0/1, 2147483647/1 ]
+      video/x-raw(memory:VASurface)
```
tjanczak: Please remove VASurface, it is deprecated memory type - replaced by VAMemory. 
VASurface and associated vaapi elements will be removed in upcoming GStreamer release.

libraries/dl-streamer/src/monolithic/gst/elements/gvagenai/genai.cpp
```diff
@@ -60,58 +63,99 @@ OpenVINOGenAIContext::~OpenVINOGenAIContext() {
...
     try {
-        GstMapInfo map;
-        if (!gst_buffer_map(buffer, &map, GST_MAP_READ)) {
-            GST_ERROR("Failed to map buffer");
-            return false;
+        // Create a GSTFrame and map to CPU memory
+        auto gst_frame = std::make_shared<dlstreamer::GSTFrame>(buffer, info);
+        auto mapped_frame = mapper->map(gst_frame, dlstreamer::AccessMode::Read);
+
+        // Convert to Mat, code from gvawatermark
```
tjanczak, line 70: Conversion of VAMemory / DMABuf to OpenCV Mat will involve memory copy. 
Please add support for remote tensors instead.

libraries/dl-streamer/docs/source/elements/gvagenai.md
yunowo: done

libraries/dl-streamer/src/monolithic/gst/elements/gvagenai/genai.cpp
yunowo: RemoteTensor is not supported. See CVS-160602

tjanczak: Is GPU memory not supported at all, or the limitation is lack of ov::preprocessor in genai front-end?
We should be able to generate RGB or BGR images in VAMemory, and create tensor wrappers around.

yunowo: Both. VAAPI surface doesn't support 24 bits RGB or BGR, only supports 32 bits RGBA or BGRA.
For GenAI, the image encoding (preprocessing) handles the data on CPU. See [classes.cpp](https://github.com/openvinotoolkit/openvino.genai/blob/releases/2025/3/src/cpp/src/visual_language/minicpm/classes.cpp) `llava_image_embed_make_with_bytes_slice` and `resample` functions.
Also the `VLMPipeline` class doesn't expose `RemoteContext`, also `RemoteContext` cannot be passed into `VLMPipeline`, so no way to create a `RemoteTensor`.

tjanczak: I see, there are lot of CPU-side memory copies in OpenVINO front-end anyway... 
As I read OV genai front-end code I believe it expects [NCHW] input tensors, and gstgenai element delivers [NHWC] input tensor - which would result in one more 'copy'. 
If OV front-end indeed expect [NCHW], we can deliver it directly from VAAPI surface.  VAAPI supports RGBP/BGRP formats (planar) which is essentially NCHW layout. 
Can you please check what input layout is expected in OV genAI front-end, and if this is indeed NCHW - can you please modify gstgenai code to deliver NCHW tensors?

tjanczak (dismissed):

tbujewsk (dismissed):

tbujewsk (approved):

PR #1065, opened 2025-10-15
support core pinning in DLSPS

xwu2intel:
### Description

Support core pinning via environment variable CORE_PINNING.
Automatic e-core, p-core, and lpe-core detection if CORE_PINNING=e-cores, p-cores, or lp-cores.

N/A

### How Has This Been Tested?

Unit tested

### Checklist:

### DIFF
--- a/microservices/dlstreamer-pipeline-server/docs/user-guide/advanced-guide/detailed_usage/how-to-advanced/performance/Core-Pinning.md
+++ b/microservices/dlstreamer-pipeline-server/docs/user-guide/advanced-guide/detailed_usage/how-to-advanced/performance/Core-Pinning.md
@@ -1,11 +1,11 @@
 
 ## Pinning the DL Streamer Pipeline Sever to CPU cores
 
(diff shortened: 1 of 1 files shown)


microservices/dlstreamer-pipeline-server/docker/run.sh
```diff
@@ -50,7 +50,18 @@ gpu_execution_prequisites
 
 ros2_prerequisites
 
-python3 -m src
+taskset_cmds=()
+case "$CORE_PINNING" in
```
tjanczak, line 54: If this is a new externally visible feature, we should document how customers can use it. 
Otherwise, we are adding a feature that only we (Intel) know how to use.

tjanczak: Let's add a User Guide section on how a customer can pin DLSPS to selected cores, and show an example how to select p-cores.

xwu2intel: added

PR #1196, opened 2025-11-07
time-series-analytics: Updated the opcua server cert name

pooja-intel:
### Description

Changes:
 src/opcua_alerts.py - Updated the secure mode connection certificate
names

Fixes # (issue)

No

### How Has This Been Tested?
Yes

### Checklist:

### DIFF
--- a/microservices/time-series-analytics/src/opcua_alerts.py
+++ b/microservices/time-series-analytics/src/opcua_alerts.py
@@ -88,13 +88,17 @@ class OpcuaAlerts:
         while attempt < max_retries:
             try:
                 if secure_mode.lower() == "true":
-                    kapacitor_cert = ("/run/secrets/"
-                                    "server_certificate.pem")
-                    kapacitor_key = ("/run/secrets/"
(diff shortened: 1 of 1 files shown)


microservices/time-series-analytics/src/opcua_alerts.py
```diff
@@ -89,9 +89,9 @@ async def connect_opcua_client(self, secure_mode, max_retries=10):
             try:
                 if secure_mode.lower() == "true":
                     kapacitor_cert = ("/run/secrets/"
-                                    "time_series_analytics_microservice_Server_server_certificate.pem")
+                                    "server_certificate.pem")
```
vkb1: As mentioned in the other PR, let's use client_certificate and client_key instead.
I'm just thinking instead of hardcoding, how about we have this as ENV vars and use it in the docker compose volume mounts the same ENVs, this would make the things more flexible.

pooja-intel: Added env for certs

PR #1248, opened 2025-11-17
Minor updates

14pankaj:
### Description

Documentation Updates for the microservice
Bug fixes
Added model type to response body

Fixes # (issue)

### How Has This Been Tested?

### Checklist:

### DIFF
--- a/libraries/dl-streamer/docs/source/dev_guide/object_tracking.md
+++ b/libraries/dl-streamer/docs/source/dev_guide/object_tracking.md
@@ -12,6 +12,7 @@ and can work in the following modes as specified by the *tracking-type* property
 | short-term-imageless | &lt;= 5                                                           | Yes                                                              | No                        | Assigns a unique id to objects and generates object position for frames on which object detection was skipped.<br>Fast algorithm that extrapolates object trajectory from previous frame(s) without access to image data. |
 | zero-term            | 1 (every frame)                                                   | Yes                                                              | Yes                       | Assigns a unique id to objects, and requires object detection run on every frame.<br>Takes into account object trajectory as well as a color histogram of object image data.                                              |
 | zero-term-imageless  | 1 (every frame)                                                   | Yes                                                              | No                        | Assigns a unique id to objects, and requires object detection run on every frame.<br>Fastest algorithm as based on comparing object coordinates on current frame with objects trajectory on previous frames.              |
+| deep-sort            | 1 (every frame)                                                   | Yes                                                              | Yes                       | Assigns a unique id to objects using Kalman filter for motion prediction and deep learning features for re-identification.<br>Robust algorithm that minimizes ID switches by combining appearance and motion cues. Requires feature extraction model (e.g., mars-small128). |
 
 ## Additional configuration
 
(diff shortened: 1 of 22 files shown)


microservices/model-download/docker/entrypoint.sh
```diff
@@ -106,49 +106,47 @@ done
...
-    IFS=',' read -ra PLUGIN_LIST <<< "$PLUGINS"
-    echo "ACTIVATED_PLUGINS=$PLUGINS" > "$PLUGINS_ENV_FILE"
-    
-    for plugin in "${PLUGIN_LIST[@]}"; do
-        install_dependencies "$plugin"
+    # Split comma-separated plugins into array and convert to lowercase
+    IFS=',' read -ra PLUGIN_LIST_RAW <<< "$PLUGINS_LOWER"
+    PLUGIN_LIST=()
+    for plugin in "${PLUGIN_LIST_RAW[@]}"; do
+        # Trim whitespace and add to array
```
bharagha, line 120: Error check not required? What if a wrong plugin or typo is there?

microservices/model-download/docker/entrypoint.sh
```diff
@@ -106,49 +106,47 @@ done
...
-    print_success "Activated plugins: $PLUGINS"
 fi
 
+# Install plugin-specific dependencies
+for plugin in "${PLUGIN_LIST[@]}"; do
+    install_dependencies "$plugin"
+done
+
+# Save activated plugins to env file (store original input)
+echo "ACTIVATED_PLUGINS=$PLUGINS" > "$PLUGINS_ENV_FILE"
```
bharagha, line 132: Why not use the lowercase version itself? Is it ensured the original input is not used anywhere else?

microservices/model-download/docker/entrypoint.sh
```diff
@@ -106,49 +106,47 @@ done
...
-if [ "$PLUGINS" = "all" ]; then
-    for plugin in "${AVAILABLE_PLUGINS[@]}"; do
-        EXTRA_ARGS+=(--extra "$plugin")
-    done
-else
-    for plugin in "${PLUGIN_LIST[@]}"; do
-        EXTRA_ARGS+=(--extra "$plugin")
-    done
-fi
+export PATH="/opt/.local/bin:$HOME/.local/bin:/opt/bin/:$PATH"
```
bharagha: Just a check that this is valid in EMT-D/EMT-S too

microservices/model-download/docker/entrypoint.sh
14pankaj: In that case the container will not start with the error as unknown plugin. Even in case on multiple comma separated plugins are passed and even one of the has type error, the container will not come up.

microservices/model-download/docker/entrypoint.sh
14pankaj: This path is valid for EMT-D and EMT-S.

14pankaj: The path has been updated as UV installation is now through PIP so no need to add it explicitly here

microservices/model-download/docker/entrypoint.sh
14pankaj: The input is being converted to lowercase and compared with the available plugin list.  Storing them in lowercase only.

bharagha (approved):

PR #1403, opened 2025-11-27
[ViPPET] Implement thread-safe access for pipeline management operations

ktaube26:
### Description

This patch introduces thread-safety to the `PipelineManager` class in the ViPPET tool by adding a shared lock to protect access to pipeline-related operations. The goal is to prevent race conditions when multiple threads interact with pipeline data.

## Key Changes
- **Added**
  - `threading.Lock` instance (`self.lock`) to `PipelineManager` for synchronized access.
- **Modified**
  - Wrapped critical sections in:
    - `add_pipeline`
    - `get_pipelines`
    - `get_pipeline_by_id`
    - `update_pipeline`
    - `delete_pipeline_by_id`
    with `with self.lock:` blocks.
- **Removed**
  - Direct, unsynchronized access to `self.pipelines` in favor of lock-protected operations.
- **Other**
  - Minor code restructuring for clarity and consistency.

## Impact
- Ensures safe concurrent modifications and reads of pipeline data.
- Reduces risk of

### Checklist:

### DIFF
--- a/tools/visual-pipeline-and-platform-evaluation-tool/vippet/managers/pipeline_manager.py
+++ b/tools/visual-pipeline-and-platform-evaluation-tool/vippet/managers/pipeline_manager.py
@@ -51,12 +51,9 @@ class PipelineManager:
         self.video_encoder = get_video_encoder()
 
     def add_pipeline(self, new_pipeline: PipelineDefinition):
-        # Check for duplicate pipeline name and version
         with self.lock:
-            if self._pipeline_exists(new_pipeline.name, new_pipeline.version):
-                raise ValueError(
-                    f"Pipeline with name '{new_pipeline.name}' and version '{new_pipeline.version}' already exists."
-                )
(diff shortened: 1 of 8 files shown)


tools/visual-pipeline-and-platform-evaluation-tool/vippet/managers/pipeline_manager.py
```diff
@@ -41,40 +43,46 @@ def get_pipeline_manager() -> "PipelineManager":
...
-
-        self.pipelines.append(pipeline)
+            self.pipelines.append(pipeline)
         self.logger.debug(f"Pipeline added: {pipeline}")
         return pipeline
 
     def get_pipelines(self) -> list[Pipeline]:
-        return self.pipelines
+        with self.lock:
+            return list(self.pipelines)
```
p-zak: Can we also use `deepcopy` here?
`return [deepcopy(p) for p in self.pipelines]` or similar

PR #1543, opened 2025-12-09
[DLStreamer] Extend latency tracer to support multiple pipelines/branches in the gst launch string 

xwu2intel:
### Description

The existing latency tracer only supports a single pipeline/branch in the gst launch string. This patch extends the support to multiple pipelines/branches. Also included is performance optimization to reduce tracing overhead. 

N/A

### How Has This Been Tested?

Unit tested with common pipelines

### Checklist:


tjanczak (commented):
Xintian, you were showing built-in latency tracers (in GStreamer) work for both single and mutli pipelines. In that case, I'd rather write a sample how to measure latency using built-pin tracers and remove DLS-streamer custom traces. This will save our maintenance effort going forward.

xwu2intel:
PR at https://github.com/open-edge-platform/dlstreamer/pull/535

PR #1572, opened 2025-12-12
[VLM] Implement telemetry metrics and performance tracking for VLM generation

bhardwaj-nakul:
### Description

- Updated FastAPI, Pydantic, and other dependencies to their latest versions in pyproject.toml.
- Enhanced model export script to handle specific model types and added error handling.
- Refactored app.py to support new model types (SmolVLM) and improved telemetry logging for request handling by enabling openvino_genai based inference for all models except SmolVLM.
- Added telemetry data models and a telemetry store for tracking usage and errors.
- Introduced utility functions for video backend selection and model video support checks.
- Updated model configuration to include new models and video support capabilities.
- Improved error handling and logging throughout the application for better observability.
- Fixed test cases

Fixes # (issue)

No

### How Has This Been Tested?

Tested in dev env

### Checklist:


microservices/vlm-openvino-serving/src/utils/telemetry.py
```diff
@@ -0,0 +1,122 @@
...
+    """Normalize a MeanStd-like object into plain floats.
+
+    Args:
+        pair: Object with ``mean`` and ``std`` attributes (e.g., ``MeanStdPair``).
+
+    Returns:
+        Tuple of ``(mean, std)`` in milliseconds/tokens/s, or (None, None) when unavailable.
+    """
+    if pair is None:
+        return None, None
```
bharagha, line 57: I know this is an internal function. But still :). What if the passed "pair" does not have the two attributes?

microservices/vlm-openvino-serving/src/utils/telemetry.py
```diff
@@ -0,0 +1,122 @@
...
+        Tuple of ``(ChatUsageStats, TelemetryMetrics)``; elements are ``None`` when the pipeline
+        did not expose metrics.
+    """
+
+    if perf_metrics is None:
+        return None, None
+
+    prompt_tokens = _safe_call(perf_metrics.get_num_input_tokens)
+    completion_tokens = _safe_call(perf_metrics.get_num_generated_tokens)
+    total_tokens = None
```
bharagha: Better to set it to 0 right just for type consistency

microservices/vlm-openvino-serving/src/utils/telemetry_store.py
```diff
@@ -0,0 +1,60 @@
...
+    """Append-only JSONL store with file locking and bounded history."""
+
+    def __init__(self, path: Path | None = None, max_records: int = _MAX_RECORDS):
+        """Configure on-disk storage location and retention budget."""
+        self.path = path or _DEFAULT_PATH
+        self.max_records = max_records
+
+    def _ensure_parent(self) -> None:
+        """Create parent directories for the telemetry file if needed."""
+        self.path.parent.mkdir(parents=True, exist_ok=True)
```
bharagha, line 27: Need to ensure that we are not using "any" folder given immutability requirements. Just a note for, say, documentation?

microservices/vlm-openvino-serving/src/utils/telemetry_store.py
```diff
@@ -0,0 +1,60 @@
...
+                lines.append(serialized)
+                if len(lines) > self.max_records:
+                    lines = lines[-self.max_records :]
+                handle.seek(0)
+                handle.truncate()
+                handle.write("\n".join(lines))
+                handle.write("\n")
+            finally:
+                fcntl.flock(handle.fileno(), fcntl.LOCK_UN)
+
```
bharagha, line 48: One question: Is the telemetry timestamped?

microservices/vlm-openvino-serving/src/utils/telemetry_store.py
```diff
@@ -0,0 +1,60 @@
...
+    def append(self, record: Dict[str, Any]) -> None:
+        """Append a telemetry record and keep only the most recent entries."""
+
+        self._ensure_parent()
+        serialized = json.dumps(record, ensure_ascii=False)
+        with open(self.path, "a+", encoding="utf-8") as handle:
+            fcntl.flock(handle.fileno(), fcntl.LOCK_EX)
+            try:
+                handle.seek(0)
+                lines = [line.rstrip("\n") for line in handle.readlines() if line.strip()]
```
bharagha, line 38: Checking: Wouldnt this read the entire file, strip and rewrite result in I/O overhead?

microservices/vlm-openvino-serving/src/utils/telemetry_store.py
```diff
@@ -0,0 +1,60 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Simple file-backed telemetry store shared across Gunicorn workers."""
+
+import fcntl
+import json
+from pathlib import Path
+from typing import Any, Dict, List
+
+_DEFAULT_PATH = Path("/tmp/vlm_telemetry.jsonl")
```
bharagha: Use /opt

microservices/vlm-openvino-serving/src/utils/telemetry_store.py
```diff
@@ -0,0 +1,60 @@
...
+_MAX_RECORDS = 100
+
+
+class TelemetryStore:
+    """Append-only JSONL store with file locking and bounded history."""
+
+    def __init__(self, path: Path | None = None, max_records: int = _MAX_RECORDS):
```
bharagha, line 20: Path | None is only with Python 3.10 onwards right?

microservices/vlm-openvino-serving/src/utils/telemetry.py
bhardwaj-nakul: `getattr` check this condition. 
getattr(pair, "std", None) return None when those attributes are missing, so the function just falls back to (None, None). Downstream code treats None as “metric unavailable”, so no exception bubbles up even if pair doesn’t expose both fields.

microservices/vlm-openvino-serving/src/utils/telemetry.py
bhardwaj-nakul: it is handled in the next line. total tokens gets updated. still for some corner case updated this to 0 instead of none.
```
if prompt_tokens is not None or completion_tokens is not None:
        total_tokens = (prompt_tokens or 0) + (completion_tokens or 0)
```

microservices/vlm-openvino-serving/src/utils/telemetry_store.py
bhardwaj-nakul: Yes, it has UTC timestamp `"timestamp": "2025-12-12T12:37:47.996Z",`

microservices/vlm-openvino-serving/src/utils/telemetry_store.py
bhardwaj-nakul: Made this field configurable from env with a default value of `/opt/vlm_telemetry.jsonl`

microservices/vlm-openvino-serving/src/utils/telemetry_store.py
bhardwaj-nakul: readlines() loads the whole file into memory, trims blanks, then rewrites every record. For the default cap (100 entries) it’s fine, but if someone cranks VLM_TELEMETRY_MAX_RECORDS into the thousands then it will result in I/O overhead.
For 100 records it will be fine as the file read write happens once per inference request. 

We can implement something like ring buffer, with a separate file maintaining a metadata of how many records have been added, and append till max and if max then delete from head without reading.

microservices/vlm-openvino-serving/src/utils/telemetry_store.py
bhardwaj-nakul: Done

microservices/vlm-openvino-serving/src/utils/telemetry_store.py
bhardwaj-nakul: Yes, it is 3.10 onwards syntax.
In pyproject.toml we are maintaining `python = "^3.10,<3.14"` which ensures that the python version is enforced.

PR #1711, opened 2026-01-23
[ViPPET] Add pipeline looping support and ExecutionConfig for flexible output modes

p-zak:
# Description

This PR introduces pipeline looping capabilities and a new unified `ExecutionConfig` model that replaces the previous `VideoOutputConfig`. The changes enable pipelines to run continuously for a specified duration while supporting different output modes.

## Key Changes

- The `VideoOutputConfig` has been replaced with `ExecutionConfig` which combines output mode selection (`disabled`, `file`, `live_stream`) with runtime control (`max_runtime`). This allows users to configure both what happens with the pipeline output and how long the pipeline should run in a single configuration object.
- A new `OutputMode` enum defines three output modes: `disabled` (default, uses fakesink), `file` (saves to file), and `live_stream` (streams to RTSP server). The `max_runtime` parameter controls how long the pipeline runs - when set to 0, pipeline runs until natural EOS; when greater than 0, pipeline loops and stops after the specified duration.
- The `Graph` class gained a new `apply_looping_modifications()` method that transforms pipelines for looping playback. It replaces `filesrc` with `multifilesrc loop=true`, changes `.mp4` extensions to `.ts`, and replaces `qtdemux` with `tsdemux` for seamless looping.
- Density tests are explicitly restricted from using `live_stream` output mode. Validation has been added in multiple places (`Benchmark.run()`, `TestsManager._validate_execution_config()`, `PipelineManager.build_pipeline_command()`) to enforce this constraint and raise `ValueError` with clear messages.
- The `VideoEncoder` class now includes low-latency streaming encoder configurations optimized for live streaming scenarios. A new `replace_fakesink_with_live_stream_output()` method handles RTSP streaming setup.
- Pipeline command building logic was updated to handle looping modifications automatically when `max_runtime > 0` and output mode is not `file`. The combination of `output_mode=file` with `max_runtime > 0` is explicitly forbidden since file output doesn't support looping.
- The GStreamer runner (`gst_runner.py`) now properly handles graceful shutdown when max-runtime is reached. A `shutdown_in_progress` flag prevents spurious error logging when elements like `rtspclientsink` report expected disconnection errors during pipeline teardown.
- All API schemas, job statuses, and manager classes have been updated to use the new `ExecutionConfig` instead of `VideoOutputConfig`. The OpenAPI specification (`vippet.json`) has been regenerated to reflect these changes.

### DIFF
--- a/tools/visual-pipeline-and-platform-evaluation-tool/vippet/managers/tests_manager.py
+++ b/tools/visual-pipeline-and-platform-evaluation-tool/vippet/managers/tests_manager.py
@@ -257,10 +257,20 @@ class TestsManager:
                 )
             )
 
-            # Update job with live_stream_urls immediately after generation
+            # Build streams distribution per pipeline
+            streams_per_pipeline = [
+                PipelinePerformanceSpec(
+                    id=spec.id,
+                    streams=spec.streams,
+                )
+                for spec in performance_request.pipeline_performance_specs
+            ]
+
+            # Update job with live_stream_urls and streams_per_pipeline immediately
             with self.lock:
                 if job_id in self.jobs:
                     job = self.jobs[job_id]
+                    job.streams_per_pipeline = streams_per_pipeline
 
                     # Type guard: ensure we have a PerformanceJob
                     if not isinstance(job, PerformanceJob):
@@ -315,20 +325,10 @@ class TestsManager:
                         job.state = TestJobState.COMPLETED
                         job.end_time = int(time.time() * 1000)
 
-                        # Build streams distribution per pipeline
-                        streams_per_pipeline = [
-                            PipelinePerformanceSpec(
-                                id=spec.id,
-                                streams=spec.streams,
-                            )
-                            for spec in performance_request.pipeline_performance_specs
-                        ]
-
                         # Update performance metrics
                         job.total_fps = result.total_fps
                         job.per_stream_fps = result.per_stream_fps
                         job.total_streams = result.num_streams
-                        job.streams_per_pipeline = streams_per_pipeline
                         job.video_output_paths = video_output_paths
 
                         self.logger.info(


tools/visual-pipeline-and-platform-evaluation-tool/vippet/managers/tests_manager.py
```diff
@@ -209,15 +250,34 @@ def _execute_performance_test(
...
+                    # Type guard: ensure we have a PerformanceJob
+                    if not isinstance(job, PerformanceJob):
+                        self.logger.error(
+                            f"Job {job_id} is not a PerformanceJob, skipping update"
+                        )
+                    else:
+                        job.live_stream_urls = live_stream_urls
+                        self.logger.debug(
+                            f"Updated job {job_id} with live_stream_urls: {live_stream_urls}"
+                        )
```
ktaube26: Can we move the part of the code (lines 318–325) that updates the job with `job.streams_per_pipeline` to this location?
Currently, `job.streams_per_pipeline` is updated when updating results after a successful pipeline execution, but in case of any failure, the job remains without any information about which pipelines were executed.

tools/visual-pipeline-and-platform-evaluation-tool/vippet/api/api_schemas.py
```diff
@@ -619,23 +619,72 @@ class PipelineRequestOptimize(BaseModel):
...
 
-    enabled: bool = Field(
-        default=False, description="Flag to enable or disable video output generation."
+    output_mode: OutputMode = Field(
+        default=OutputMode.DISABLED,
+        description="Mode for pipeline output generation.",
+    )
+    max_runtime: float = Field(
+        default=0.0,
+        ge=0.0,
```
ktaube26, line 686: Why is the type of `max_runtime` a float instead of int?

tools/visual-pipeline-and-platform-evaluation-tool/vippet/managers/tests_manager.py
p-zak: Done

tools/visual-pipeline-and-platform-evaluation-tool/vippet/api/api_schemas.py
p-zak: By design but you can pass `int` if you want, it will work perfectly fine.

PR #1728, opened 2026-01-29
[Document Ingestion] Updated SSRF logic

Supriya-Krishnamurthi:
### Description

This PR includes below changes:

- Refactored code to remove pinning url logic as it leads to ssl verification failure on resolved pinned IP
- I have added all possible SSRF checks. Any related warning is false positive
- Fixed unintended URL re-fetching by refactoring parse_html to process already-fetched HTML, enforcing a single fetch per URL.
  - *parse_html* was re-fetching the URL independently, bypassing the URL validation and safe_fetch_url path. This created a DNS rebinding risk, where the initial fetch could resolve to a public IP and pass validation, but a subsequent fetch could resolve to a private or loopback IP without re-validation
- Removed redundant URL-processing loops.
- Extended public IP validation from IPv4-only to both IPv4 and IPv6 by validating all resolved addresses.
- Improved ingestion behavior to allow partial success, failing only when all URLs fail and logging per-URL failures.
- Added ALLOWED_HOSTS to Docker environment configuration
- Enforced SSL verification during safe URL fetch

Fixes # https://jira.devtools.intel.com/browse/ITEP-83476

### How Has This Been Tested?

### Checklist:

### DIFF
--- a/microservices/document-ingestion/pgvector/app/url.py
+++ b/microservices/document-ingestion/pgvector/app/url.py
@@ -90,7 +90,8 @@ def validate_url(url: str) -> bool:
 
         # Parse the cleaned URL
         parsed_url = urlparse(url_cleaned)
-        if parsed_url.scheme not in ["http", "https"]:
+        # Ensure the URL are secure (https only)
+        if parsed_url.scheme != "https":
             return False
 
         hostname = parsed_url.hostname
--- a/microservices/document-ingestion/pgvector/docs/user-guide/get-started.md
+++ b/microservices/document-ingestion/pgvector/docs/user-guide/get-started.md
@@ -273,6 +273,8 @@ commands from the same shell as where the environment variables are set.
  pages (social media feeds, Single Page Applications), the API may indicate a successful request
  but the actual content might not be captured. Such pages should be avoided or handled separately.
 
+ > **Security Note**: This microservice only allows secure URLs using the HTTPS protocol.
+
 1. **Upload URLs to create and store embeddings**:
    Submit one or more URLs to be processed for embedding creation.
    ```bash
--- a/sample-applications/chat-question-and-answer/docs/user-guide/get-started.md
+++ b/sample-applications/chat-question-and-answer/docs/user-guide/get-started.md
@@ -160,9 +160,11 @@ Visit https://huggingface.co/settings/tokens to get your token.
 
 7. **Access the Application**:
    Open a browser and go to `http://<host-ip>:8101` to access the application dashboard. The application dashboard allows the user to,
-    - Create and manage context by adding documents (pdf, docx, etc.) and web links. Note: There are restrictions on the max size of the document allowed.
+    - Create and manage context by adding documents (pdf, docx, etc.) and web links (only https).
+    > **Note**:
+    > - There are restrictions on the max size of the document allowed.
+    > - Only HTTPS URLs are allowed for web links.
     - Start Q&A session with the created context.
-
 
 ## Running in Kubernetes
 


microservices/document-ingestion/pgvector/app/main.py
```diff
@@ -367,10 +367,13 @@ async def ingest_links(urls: list[str]) -> dict:
     """
     try:
         if urls:
-            ingest_url_to_pgvector(urls)
+            result = ingest_url_to_pgvector(urls)
+
+        return {
```
bharagha, line 377: What if URL is empty? Return is not yet initialized right. Will it not be a critical issue?

microservices/document-ingestion/pgvector/app/main.py
```diff
@@ -367,10 +367,13 @@ async def ingest_links(urls: list[str]) -> dict:
     """
     try:
         if urls:
-            ingest_url_to_pgvector(urls)
+            result = ingest_url_to_pgvector(urls)
```
bharagha: async function, but ingest can be sync. Isnt it? Do you want to add "await"?

microservices/document-ingestion/pgvector/app/main.py
```diff
@@ -367,10 +367,13 @@ async def ingest_links(urls: list[str]) -> dict:
...
+
+        return {
+            "status": 200,
+            "message": f"Data preparation completed: {result['successful']}/{result['total_urls']} URLs succeeded",
+        }
```
bharagha, line 380: Will this work all the time? What if none is returned?

microservices/document-ingestion/pgvector/app/main.py
```diff
@@ -367,10 +367,13 @@ async def ingest_links(urls: list[str]) -> dict:
...
 
-        result = {"status": 200, "message": "Data preparation succeeded"}
-        return result
 
     except HTTPException as e:
         raise e
```
bharagha: Just raise is sufficient as you are not changing anything...

microservices/document-ingestion/pgvector/app/url.py
```diff
@@ -72,43 +72,16 @@ def is_public_ip(ip: str) -> bool:
     except ValueError:
         return False
 
-def construct_pinned_url(parsed_url, resolved_ip):
-    """
-    Constructs a pinned URL using the resolved IP address and the original parsed URL components.
-
-    Args:
-        parsed_url (ParseResult): The parsed URL object.
-        resolved_ip (str): The resolved IP address to use in the netloc.
```
bharagha, line 76: I dont fully understand the new logic w.r.t SSRF protection. As I understand, previously the URL was getting pinned. Now you resolve once but DNS resolution is happening on its own later. So, DNS is changed by attacker, potentially SSRF protection fails, isnt it?

microservices/document-ingestion/pgvector/app/url.py
```diff
@@ -117,66 +90,58 @@ def validate_url(url: str) -> Optional[str]:
...
+            return False
+
+        # Check if hostname matches allowed domains (exact or subdomain match)
+        if not any(normalized_hostname == allowed or normalized_hostname.endswith('.' + allowed) 
+                   for allowed in allowed_domains):
+            logger.info(f"URL hostname {normalized_hostname} is not in the whitelisted domains {allowed_domains}.")
+            return False
+
+        # Resolve ALL IPs for the hostname
+        infos = socket.getaddrinfo(hostname, None)
```
bharagha: Not normalized hostname? Also, not catching the error...

microservices/document-ingestion/pgvector/app/url.py
```diff
@@ -186,130 +151,124 @@ def safe_fetch_url(validated_pinned_url: str, headers: dict, hostname: str):
...
-
     # Send the request to the validated URL to prevent SSRF
     response = requests.get(
-        validated_pinned_url,
+        validated_url,
         headers=headers,
         timeout=5,
-        allow_redirects=False  # prevent redirect SSRF
+        allow_redirects=False,  # prevent redirect SSRF
+        verify=True  # enforce SSL verification
```
bharagha, line 165: There is a critical error raised from code scanning...

microservices/document-ingestion/pgvector/app/utils.py
```diff
@@ -45,47 +44,24 @@ def get_separators():
...
-def load_html_content(links, chunk_size=1500, chunk_overlap=50):
-    """
-    Loads HTML content from the provided links, processes it into text, and splits it into chunks.
-    Args:
-        links (list): A list of URLs to load HTML content from.
-        chunk_size (int, optional): The maximum size of each text chunk. Defaults to 1500.
-        chunk_overlap (int, optional): The number of overlapping characters between consecutive chunks. Defaults to 50.
-    Returns:
-        list: A list of processed and split text documents.
+def parse_html_content(html_content: str, source_url: str = "") -> str:
```
bharagha, line 47: Changes in this file have resulted in modified behavior. Have you checked that this does not break anything?

Supriya-Krishnamurthi: Previously prase_html()  fetched the URL again exposing to several security issues as second fetch bypassed all validation (domain whitelist, IP checks, redirect blocking). This current function ensures one validated fetch per URL where it parses already fetched URL content and converts it to plain text and doesnt break anything. The refactored function was only used in url_ingestion

microservices/document-ingestion/pgvector/app/url.py
Supriya-Krishnamurthi: CodeQL is flagging a potential SSRF because it cannot see that url is validated. With pinned URL logic removed, the warning appears, but current safeguards—redirects blocked, whitelisted domains, and public IP checks—already minimize the risk. Once we decide whether to validate+fetch in the same function or retain pinned URL with proper SSL handling as mentioned in previous comment, I can update the implementation to resolve the CodeQL issue.

microservices/document-ingestion/pgvector/app/url.py
Supriya-Krishnamurthi: Right! These were present in the original code but got missed during refactoring rom IPv4-only (socket.gethostbyname()) to IPv4+IPv6 resolution (socket.getaddrinfo()). Fixed now

microservices/document-ingestion/pgvector/app/url.py
Supriya-Krishnamurthi: Valid point, Pinning URL did help overcome DNS rebinding issue but introduces SSL error:

### Original URL
https://example.com/page

### Pinned to IP
https://192.0.2.1/page  # SSL cert is for "example.com", not the IP!

Considering the  current safeguards, with not allowing redirects, allowing whitelisted domains, public IP verification, the risk is lower hence updated the logic.  But here are 2 suggestions that I came across: 

1: In the same function do both validation and fetch , so that it minimizes window to milisec between DNS validation and request

2: If we have to retain pinned url logic, we should handle ssl issue better. Because disabling ssl verification defeats security purpose. To maintain both, Options include to write custom DNS resolver to pin IP with hostname preservation or custom SNI support. 

Need your inputs on this or suggest if you have a better approach

microservices/document-ingestion/pgvector/app/main.py
Supriya-Krishnamurthi: Right. I noticed several endpoints use redundant exception patterns (e.g. raise e) in multiple places. I’ve cleaned this up in the functions I touched. We might have to standardize exception handling across the module in a follow-up PR

microservices/document-ingestion/pgvector/app/main.py
Supriya-Krishnamurthi: If url_list is empty:

Loop never executes, so
invalid_urls = 0
len(url_list) = 0
if invalid_urls == len(url_list):
        raise HTTPException()
Check: 0 == 0 → True → raises exception 

So the current code cannot return None. Also raised exception in the endpoint incase of empty url list

microservices/document-ingestion/pgvector/app/main.py
Supriya-Krishnamurthi: Since it is sync, we dont need await

microservices/document-ingestion/pgvector/app/main.py
Supriya-Krishnamurthi: raised an exception incase of empty url list

microservices/document-ingestion/pgvector/app/url.py
bharagha: Can you drop a note to SDL/Security team?

Supriya-Krishnamurthi: Okay sure, I will also tag Umesh here @ucjaiswal

ucjaiswal: @Supriya-Krishnamurthi, as discussed, since you are already doing domain allowlisting, we are fine with this. We generally trust fully qualified domain names as IP addresses may change. With URLs pinning/domain allowlisting with blocking localhost & internal/private host/IPs we are good from SSRF perspective. Just ensure that we also remove "http" , Public access over http is vulnerable to MiMt (man-in-the-middle) attack which can lead to similar consequences like SSRF.

PR #1922, opened 2026-03-09
feat: upgrade URL ingestion to use MarkdownHeaderTextSplitter

ishaanv1709:
Resolves #1921 
@bharagha @14pankaj @jgespino @krish918 
# Description
This PR upgrades the chunking pipeline in [url.py](cci:7://file:///c:/Users/ishaa_04bpft8/Desktop/Openvivo%20GSOC/edge-ai-libraries/microservices/document-ingestion/pgvector/app/url.py:0:0-0:0) from a character-based approach to a structural one.

- **Problem:** Websites converted to Markdown often contain code blocks or tables. Using a character splitter blindly slices through these, destroying context for the LLM.
- **Solution:** Implemented a two-stage process using `MarkdownHeaderTextSplitter`. It first groups content by headers (`#`, `##`), then applies token limits. This ensures structural elements (like Python scripts) stay intact.
- **Scope:** Narrowed strictly to Web Ingestion. `.docx` and `.pdf` pipelines in [document.py](cci:7://file:///c:/Users/ishaa_04bpft8/Desktop/Openvivo%20GSOC/edge-ai-libraries/microservices/document-ingestion/pgvector/app/document.py:0:0-0:0) remain unchanged as they lack native Markdown headers.

# Testing
Verified via local string-splitting evaluations using a pytest evaluation. The previous logic shattered code blocks across chunks; the new logic successfully retains them as unified semantic units.

Screenshot 2026-03-09 174651


bharagha:
Thanks for the response. I have reviewed the code and see no issues with bringing it in. We will check pulling this specific PR to the mainline. 

My question on why this PR is relevant was mainly in the context of GraphRAG problem statement. Let's ensure that cycles are more spent on that problem statement. I also looped in @naitik-2006 on this topic to collaborate with you.

microservices/document-ingestion/pgvector/app/url.py
```diff
@@ -186,7 +186,17 @@ def ingest_url_to_pgvector(url_list: List[str]) -> dict:
     }
 
     # Initialize text splitter and embedder once
-    text_splitter = RecursiveCharacterTextSplitter(
+    headers_to_split_on = [
```
Supriya-Krishnamurthi, line 189: The implementation assumes that all ingested content contains Markdown headers. In the current pipeline, content is derived from HTML and may not always preserve header structure in Markdown form. In such cases, it falls back to existing character-based splitting. Code doesnt break for non-md files but doesnt provide much benefits in such cases.

And when no headers are present, this fallback is implicit and not transparent. It would be better to explicitly detect the presence of structure and fallback to RecursiveCharacterTextSplitter when headers are not found, to make the behavior more predictable?

Since this is a sample application, should we  introduce a config flag (e.g., enabling/disabling structured chunking) to keep the default (can be defaulted to disable) pipeline simple and easier to reason about. User can enable when sure about retrieved content being structured

microservices/document-ingestion/pgvector/app/url.py
```diff
@@ -186,7 +186,17 @@ def ingest_url_to_pgvector(url_list: List[str]) -> dict:
...
+        ("#", "Header 1"),
+        ("##", "Header 2"),
+        ("###", "Header 3"),
+        ("####", "Header 4"),
+    ]
+    markdown_splitter = MarkdownHeaderTextSplitter(
```
Supriya-Krishnamurthi, line 195: Header-based splitting alone may not fully preserve semantic units like **code blocks or tables**. Not a blocker but we will still miss out on context for other such structured components in the content which may not fully align with the intent of this PR

microservices/document-ingestion/pgvector/app/url.py
```diff
@@ -227,13 +237,18 @@ def ingest_url_to_pgvector(url_list: List[str]) -> dict:
                 continue
 
             # Chunk + embed
-            chunks = text_splitter.split_text(content)
-            metadata = [{"url": url}] * len(chunks)
+            md_header_splits = markdown_splitter.split_text(content)
```
Supriya-Krishnamurthi, line 240: The hierarchical relationship between headers may not be fully preserved. While content is grouped under headers, the linkage between parent and child sections (e.g., subheaders B, C under A and E, F under D) is not explicitly maintained beyond individual chunks. Although MarkdownHeaderTextSplitter may initially capture this hierarchy in metadata, it can be lost after applying RecursiveCharacterTextSplitter and subsequently updating metadata (e.g., doc.metadata["url"] = url), which may overwrite existing fields. Request to validate such scenarios and share the repot

microservices/document-ingestion/pgvector/app/url.py
```diff
@@ -227,13 +237,18 @@ def ingest_url_to_pgvector(url_list: List[str]) -> dict:
...
+            docs = char_splitter.split_documents(md_header_splits)
+            
+            for doc in docs:
+                doc.metadata["url"] = url
```
Supriya-Krishnamurthi, line 244: Double check on the metadata handling here as the metadata from markdown-split can be overwritten. May be after checking flag for structured content splitting or text splitting, we can retain this metadata accordingly.

microservices/document-ingestion/pgvector/app/url.py
```diff
@@ -227,13 +237,18 @@ def ingest_url_to_pgvector(url_list: List[str]) -> dict:
                 continue
 
             # Chunk + embed
-            chunks = text_splitter.split_text(content)
-            metadata = [{"url": url}] * len(chunks)
+            md_header_splits = markdown_splitter.split_text(content)
+            docs = char_splitter.split_documents(md_header_splits)
```
Supriya-Krishnamurthi, line 241: Just checking, if large amounts of content exist under a single header without any sub-headers, the subsequent character-based splitting can still split the large context in a similar way as the current approach

bharagha:
@ishaanv1709 : Any feedback to the comments?

PR #2101, opened 2026-04-10
[Model Download] Upgrade dlstreamer script version to v2026.0.0 for public model downloads and support INT8 quantization

Supriya-Krishnamurthi:
#### Summary

- Upgraded download_public_models.sh from DL Streamer to v2026.0.0 and added support for INT8 quantization in the Ultralytics plugin.
- Extended model download to support additional models enabled by the updated script.
- Updated openapi.yaml to include the quantize field in the API schema.

#### Observed Limitation (Upstream)

- Identified limitations in the DL Streamer script where INT8 requests are not reliably validated.
- When a user explicitly requests INT8 for an unsupported model, the script may:
    - Succeed without generating INT8 artifacts and silently fall back to FP16/FP32
    - Provide no clear indication of INT8 conversion failure
    - The documentation does list the models that support int 8 but not handled in the script
- Find more details on the issue via: https://github.com/open-edge-platform/dlstreamer/issues/755

#### Enhancements / Fixes

To handle the above limitation on the Model Download side:

- Added validation to ensure INT8 artifacts are generated during download
- Introduced fail-fast behavior:
    - If INT8 is requested but not produced, artifacts are cleaned up and a clear error is raised

#### Test Coverage

Expanded unit tests in test_ultralytics_plugin.py to cover:

- Script failure during INT8 requests
- Successful execution with missing INT8 artifacts
- Rejection of invalid inputs (comma-separated, all, yolo_all) when quantize is set
   Note: Restricting to a single model for INT8 avoids ambiguity, since the current DL Streamer script does not clearly report which model failed quantization in multi-model requests. This can be revisited once the upstream issue is resolved.

Fixes # (issue)

### How Has This Been Tested?

### Checklist:


microservices/model-download/src/plugins/ultralytics_plugin.py
```diff
@@ -40,15 +41,44 @@ def download(self, model_name: str, output_dir: str, **kwargs) -> Dict[str, Any]
         #model_without_prefix = model_name.split(":")[-1] if ":" in model_name else model_name
         
         # Extract quantization from kwargs
-        quantize = kwargs.get("quantize", "")
+        quantize = kwargs.get("quantize", "").strip()
+        int8_requested = bool(quantize)
+
+        # Validate: INT8 quantization requires single model (not comma-separated, all, or yolo_all)
+        if int8_requested:
```
hteeyeoh, line 52: This Int8 validation will be a temperory workaround until they fix it? Just curious.

Supriya-Krishnamurthi: yes restricting users to pass specific model instead of 'all' or 'comma sepearted' incase on INT8 request is temp fix. Otherwise to handle that will be over engineering to check every downloaded model folder for INT8 artifacts presence and then cleanup that folder if not present

hteeyeoh: ok got it.

microservices/model-download/src/api/models.py
```diff
@@ -286,6 +286,7 @@ class ModelRequest(BaseModel):
     type: Optional[ModelType] = None
     is_ovms: bool = False
     revision: Optional[str] = None
+    quantize: Optional[str] = None # Ultralytics quantization dataset (for example: coco) used to enable INT8 export
```
14pankaj: Keep this inside the config object. Instead a separate parameter as this will allow to keep the any model config to be at one object.

14pankaj (changes requested):

microservices/model-download/src/api/models.py
Supriya-Krishnamurthi: updated

bharagha (approved):

14pankaj (approved):

PR #2141, opened 2026-04-23
[ViPPET] Created gvagenai pipeline

Supriya-Krishnamurthi:
## Summary

This PR adds the GVAGenAI-based Video Summarization VLM pipeline in ViPPET. It includes backend fixes for GVAGenAI execution, pipeline-template updates for correct model and metadata handling, and UI updates for configuring and displaying the new node cleanly.

### 1. Prevent `decodebin3` replacement for GVAGenAI pipelines
The existing backend logic replaced `decodebin3` with `parsebin ! avdec_h264` before execution. That worked for normal video pipelines, but it caused some linking failures for GVAGenAI.

This change adds a GVAGenAI-aware guard so those pipelines keep `decodebin3` unchanged. As a result, the video summarization pipeline no longer fails with the `qtdemux not-linked`  startup issue.

### 2. Fix output placeholder handling for metadata-only pipelines
`gvagenai` produces metadata-only output, so the final `fakesink` in the summarization pipeline is only used to terminate the pipeline after JSON metadata is written. Previously, the backend treated any pipeline with a single `fakesink` as the main video output placeholder.

That assumption was dint work for GVAGenAI pipelines and caused unnecessary video-output injection logic to run. This change adds a GVAGenAI-aware check so metadata-only pipelines are handled correctly and are no longer misclassified as normal video-output pipelines.

### 3. Persist generated summaries instead of discarding them
The original video summarization pipeline published metadata to `/dev/null`, which meant GVAGenAI summaries were generated but immediately discarded.

This change updates the pipeline template to write JSON Lines metadata to `/videos/output/summary.jsonl`. The generated scene summaries are now preserved and can be inspected after the pipeline finishes. 

_**Need to confirm if this is the right path to store the output json**_

### 4. Model 
Currently manually downloaded Gemma3 INT4 from stand-alone model-download service and placed it in relavant folder and updated supported_models.yaml with path.

_**Post integration with Model download in vippet, need to re-visit this part**_

### 5. Add GVAGenAI-specific UI configuration fields
The GVAGenAI node needed UI-level configuration for prompt, generation config, frame rate, chunk size, and metrics. So users can inspect and modify GVAGenAI properties from the pipeline editor. The YAML still provides defaults, while the UI config makes those values editable.

**_Edges are properly connected to the middle of every node, but the gvagenai node appears with a left offset. Need to check on this_**

### 6. GVAGenAI node display in the pipeline editor
The initial commit GVAGenAI node display had a few usability issues, so removed unnecessary prompt text from the compact display etc.

### 7. Add FPS measurement for the summarization pipeline
The initial commit in summarization pipeline did not include `gvafpscounter`, so runtime fps could not be measured in the same way as other reference pipelines.

This change adds `gvafpscounter starting-frame=100` so FPS is reported after warmup rather than during initial model setup, similar to other pipelines. 

### 8. Add test coverage for GVAGenAI sink handling
The output-placeholder bug was subtle because it only affected metadata-only pipelines that happened to end in a single unnamed `fakesink`. It ensures GVAGenAI pipelines with a single unnamed `fakesink` are no longer converted into normal video-output placeholder flows.

## Validation

- Verified the CPU Video Summarization VLM pipeline runs end-to-end with GVAGenAI.
- Confirmed summaries are written to `shared/videos/output/summary.jsonl`.
- Confirmed FPS is reported after adding `gvafpscounter`.
- Added test coverage for GVAGenAI metadata-only sink handling.

Fixes # (issue)

### How Has This Been Tested?

### Checklist:


tools/visual-pipeline-and-platform-evaluation-tool/ui/src/features/pipeline-editor/nodes/GVAGenAINode.tsx
```diff
@@ -0,0 +1,95 @@
...
+    "chunk-size"?: string;
+    prompt?: string;
+    "generation-config"?: string;
+    metrics?: string;
+  };
+};
+
+const GVAGenAINode = ({ data }: GVAGenAINodeProps) => {
+  const { simpleGraph } = usePipelineEditorContext();
+  const modelValue = data.model || "";
```
dlemiech: use Nullish Coalescing '??'

tools/visual-pipeline-and-platform-evaluation-tool/vippet/pipelines/images/video-summary.png
KaliszWiktoria, line 1: The image seems to be distinctly darker than the rest of them:
 
Image

tools/visual-pipeline-and-platform-evaluation-tool/vippet/pipelines/video-summarization-vlm.yaml
```diff
@@ -0,0 +1,58 @@
+name: Video Summarization VLM
+definition: >
+  Video summarization pipeline using gvagenai with a vision-language model to generate concise scene-level summaries from sampled frames.
+tags:
+  - genAi
```
KaliszWiktoria: Shouldn't we make genAi tag begin with a capital letter like the rest?

Image

tools/visual-pipeline-and-platform-evaluation-tool/vippet/pipelines/images/video-summary.png
Supriya-Krishnamurthi: Yes, this one was created by me for POC, can you share the official image that I can use here

tools/visual-pipeline-and-platform-evaluation-tool/vippet/api/api_schemas.py
```diff
@@ -148,20 +148,22 @@ class DeviceFamily(str, Enum):
...
-    **Model category for classification or detection tasks.**
+    **Model category for classification, detection, or GenAI tasks.**
 
-    ## Values
-    - `CLASSIFICATION` - Classification model
-    - `DETECTION` - Detection model
+      ## Values
+      - `CLASSIFICATION` - Classification model
+      - `DETECTION` - Detection model
+    - `GENAI` - Generative AI model (for example VLM)
```
p-zak:
```suggestion
    ## Values
    - `CLASSIFICATION` - Classification model
    - `DETECTION` - Detection model
    - `GENAI` - Generative AI model (for example VLM)
```

tools/visual-pipeline-and-platform-evaluation-tool/vippet/api/api_schemas.py
```diff
@@ -148,20 +148,22 @@ class DeviceFamily(str, Enum):
...
 
-    ### Example
-    ```json
-    "detection"
-    ```
+      ### Example
+      ```json
+      "detection"
+      ```
```
p-zak:
```suggestion
    ### Example
    ```json
    "detection"
    ```
```

tools/visual-pipeline-and-platform-evaluation-tool/ui/src/features/pipeline-editor/nodes/GVAGenAINode.tsx
Supriya-Krishnamurthi: updated

tools/visual-pipeline-and-platform-evaluation-tool/vippet/pipelines/images/video-summary.png
Supriya-Krishnamurthi: updated new image

tools/visual-pipeline-and-platform-evaluation-tool/vippet/pipelines/video-summarization-vlm.yaml
Supriya-Krishnamurthi: Done

tools/visual-pipeline-and-platform-evaluation-tool/vippet/graph.py
```diff
@@ -603,13 +606,18 @@ def prepare_main_output_placeholder(self) -> "Graph":
...
                 node = fakesink_nodes[0]
                 node.data.clear()
                 node.type = OUTPUT_PLACEHOLDER
                 placeholder_created = True
                 logger.debug(f"Converted node {node.id} to OUTPUT_PLACEHOLDER")
+            elif has_gvagenai:
+                # gvagenai pipeline - metadata-only, keep unnamed fakesink as-is
+                logger.debug(
+                    "gvagenai pipeline detected. Keeping unnamed fakesink for metadata output."
+                )
```
p-zak: Two concerns with this block:

1. **Coupling to a specific element name.** The rule "metadata-only pipeline → keep unnamed fakesink as-is" is expressed as `node.type == "gvagenai"`. Any future GenAI-style element will require editing this function again, plus the matching place in `pipeline_manager.py`.

2. **Duplicated check.** The same `has_gvagenai = any(... == "gvagenai" ...)` appears in `pipeline_manager.build_pipeline_command` (~line 688). The two have to stay in sync - if someone later renames the element or adds a second one, missing one of the two sites silently produces wrong behavior (either a misleading "missing output sink" error or an unintended placeholder conversion).

Suggestion: introduce a single source of truth at module level and reuse it everywhere:

```python
# graph.py (top of file, near OUTPUT_PLACEHOLDER)
METADATA_ONLY_NODE_TYPES: frozenset[str] = frozenset({"gvagenai"})


def graph_is_metadata_only(nodes: list[Node]) -> bool:
    """Return True if the pipeline produces only metadata (no video output)."""
    return any(node.type in METADATA_ONLY_NODE_TYPES for node in nodes)
```

Then this block becomes:

```python
if not placeholder_created:
    is_metadata_only = graph_is_metadata_only(modified_graph.nodes)
    fakesink_nodes = [n for n in modified_graph.nodes if n.type == "fakesink"]

    if not fakesink_nodes:
        raise ValueError(...)
    elif is_metadata_only:
        logger.debug(
            "Metadata-only pipeline detected. Keeping unnamed fakesink as-is."
        )
    elif len(fakesink_nodes) == 1:
        node = fakesink_nodes[0]
        node.data.clear()
        node.type = OUTPUT_PLACEHOLDER
        placeholder_created = True
        logger.debug(f"Converted node {node.id} to OUTPUT_PLACEHOLDER")
    else:
        raise ValueError(...)
```

`pipeline_manager.py` then calls the same `graph_is_metadata_only(...)` so the two locations cannot drift apart.

tools/visual-pipeline-and-platform-evaluation-tool/vippet/graph.py
```diff
@@ -2568,7 +2577,10 @@ def _build_chain(
             # Regular element: type followed by space-separated properties.
             result_parts.append(node.type)
             for key, value in node.data.items():
-                result_parts.append(f"{key}={value}")
+                output_key = (
+                    "model-path" if node.type == "gvagenai" and key == "model" else key
+                )
+                result_parts.append(f"{output_key}={value}")
```
p-zak, line 2595: Hardcoding `node.type == "gvagenai" and key == "model"` to remap to `model-path` couples the generic chain renderer to a single element. There's also an inverse mapping in `_model_path_to_display_name` (lines 2641–2661) and a "skip model_proc" exception in `_model_display_name_to_path` (line 2718). That's three places that must stay in lockstep for one element.

Suggestion: define a small per-node-type config table near the top of the file and use it everywhere:

```python
# Per-node-type settings for elements that deviate from the default
# `data["model"]` convention. Default for any other type is:
#   model_key="model", uses_model_proc=True
@dataclass(frozen=True)
class NodeModelSpec:
    model_key: str = "model"          # Key used in the GStreamer string
    uses_model_proc: bool = True      # Whether to inject model-proc after model

NODE_MODEL_SPECS: dict[str, NodeModelSpec] = {
    "gvagenai": NodeModelSpec(model_key="model-path", uses_model_proc=False),
}

def _model_spec(node_type: str) -> NodeModelSpec:
    return NODE_MODEL_SPECS.get(node_type, NodeModelSpec())
```

Then the three call sites become:

```python
# _build_chain
spec = _model_spec(node.type)
for key, value in node.data.items():
    output_key = spec.model_key if key == "model" else key
    result_parts.append(f"{output_key}={value}")
```

```python
# _model_path_to_display_name
spec = _model_spec(node.type)
model_path = node.data.get(spec.model_key)
...
node.data["model"] = display_name_or_empty
if spec.model_key != "model":
    node.data.pop(spec.model_key, None)
```

```python
# _model_display_name_to_path
spec = _model_spec(node.type)
node.data["model"] = model.model_path_full
if spec.uses_model_proc and model.model_proc_full:
    _insert_model_proc_after_model(node, model.model_proc_full)
```

Same external behavior, but each new node type is one entry in `NODE_MODEL_SPECS` instead of N scattered `if`s.

tools/visual-pipeline-and-platform-evaluation-tool/vippet/graph.py
```diff
@@ -2642,12 +2655,17 @@ def _model_path_to_display_name(nodes: list[Node]) -> None:
         )
 
         if model is not None:
+            # Use "model" as canonical key in graph/UI for all node types.
             node.data["model"] = model.display_name
+            if model_key != "model":
+                node.data.pop(model_key, None)
```
p-zak, line 2673: Note the asymmetric handling: input may carry either `model` or `model-path` (depending on `node.type`), but the output always sets `node.data["model"]` and pops `model-path`. That's fine as a normalization pass, but it means the canonical in-memory representation differs from what `_build_chain` later emits. This is an extra reason to centralize the convention as suggested in the comment on `_build_chain` - otherwise readers have to mentally reconstruct: "key in YAML/string is `model-path`, key in graph is `model`, key emitted again is `model-path`".

If you adopt the `NodeModelSpec` table, the docstring of `_model_path_to_display_name` should also explicitly state: "the canonical in-memory key is always `model`, regardless of how the source pipeline string named it".

tools/visual-pipeline-and-platform-evaluation-tool/vippet/graph.py
```diff
@@ -2256,8 +2264,9 @@ def _tokenize(element: str) -> Iterator[_Token]:
         Output: [Token(TYPE, "filesrc"), Token(PROPERTY, "location=/tmp/foo.mp4")]
     """
     token_specification = [
-        # Property in key=value format (no commas here; caps are handled separately)
-        ("PROPERTY", r"\S+\s*=\s*\S+"),
+        # Property in key=value format, with support for quoted values containing spaces
+        # Matches: key=value, key="quoted value", key='quoted value'
+        ("PROPERTY", r'\S+\s*=\s*(?:"[^"]*"|\'[^\']*\'|\S+)'),
```
p-zak, line 2281: The new regex `\S+\s*=\s*(?:"[^"]*"|\'[^\']*\'|\S+)` is a useful extension, but two things are worth handling/testing:

1. **Escaped quotes** are not supported. A value like `prompt="say \"hi\""` would be tokenized incorrectly (matched up to the first inner `"`). That's not used in this PR, but as soon as a prompt contains a quote (very plausible for VLMs), it will break silently. If you don't want to handle escapes now, please add a TODO and a clear error if a quoted string is unterminated.
2. **No tests**. The tokenizer is a critical, generic primitive - it is now also used by all existing `gva*` elements. Please add explicit tests in `graph_test.py` for:
   - `key="value with spaces"`
   - `key='value with spaces'`
   - mixed: `a=1 b="x y" c=3`
   - round-trip through `_build_chain`: parse a pipeline string, re-emit it, parse again - the second result must equal the first.

Suggested regex with a small improvement (non-greedy + escape support, optional):

```python
("PROPERTY", r'\S+\s*=\s*(?:"(?:\\.|[^"\\])*"|\'(?:\\.|[^\'\\])*\'|\S+)'),
```

This handles `"a \"b\" c"` and `'a \'b\' c'`. If you go this route, mirror the unescaping in `_build_chain` so the round-trip stays correct.

tools/visual-pipeline-and-platform-evaluation-tool/vippet/graph.py
```diff
@@ -2697,7 +2715,7 @@ def _model_display_name_to_path(nodes: list[Node]) -> None:
 
         node.data["model"] = model.model_path_full
 
-        if model.model_proc_full:
+        if node.type != "gvagenai" and model.model_proc_full:
```
p-zak, line 2730: `if node.type != "gvagenai" and model.model_proc_full:` - same coupling concern as above. With the `NodeModelSpec` table this becomes `if spec.uses_model_proc and model.model_proc_full:` which reads as a property of the node type rather than a hardcoded exception.

tools/visual-pipeline-and-platform-evaluation-tool/vippet/managers/pipeline_manager.py
```diff
@@ -683,26 +683,35 @@ def build_pipeline_command(
...
-                        raise ValueError(
-                            "Output subpipeline was not created as expected."
+                        if not has_gvagenai:
+                            raise ValueError(
+                                f"Pipeline '{pipeline_name}' (id: {pipeline_id}) is missing required output sink. "
+                                f"Please add 'fakesink name=default_output_sink' at the end of the pipeline definition."
+                            )
+                        logger.debug(
+                            "gvagenai pipeline detected with unnamed fakesink. Skipping output injection."
+                        )
```
p-zak: Three concerns with this block:

1. **Duplicated logic.** `has_gvagenai = any(n.type == "gvagenai" ...)` is exactly the same heuristic as in `graph.prepare_main_output_placeholder`. If/when a second metadata-only element is added, this place will be missed. Please reuse a shared helper (see comment on `graph.py` ~line 595 - `graph_is_metadata_only(...)`).

2. **Silent skip when `output_mode != DISABLED`.** If a user picks "Save to file" / "Live stream" for a gvagenai pipeline, the code logs a debug message and just doesn't inject the output sink. The user will see a successful run with no output and no warning. This contract should be explicit at a higher level.

3. **Test coverage.** There's a graph-level test for `prepare_main_output_placeholder`, but no test in `pipeline_manager` exercising the new `if not has_gvagenai → raise / else log` branch. Please add one (parametrize over `output_mode` values).

tools/visual-pipeline-and-platform-evaluation-tool/vippet/models.py
```diff
@@ -52,7 +52,7 @@ def __init__(
         self.display_name: str = display_name
         self.source: str = source
         self.model_type: str = model_type
-        self.model_path: str = model_path
+        self.model_path: str = os.path.normpath(model_path)
```
p-zak, line 55: Can we have additional comment here?

For example:
```
# `model_path` is normalized at construction time so that comparisons in
# `find_installed_model_by_model_and_proc_path` are stable across YAML
# entries with trailing slashes, redundant separators, or `./` segments.
```

tools/visual-pipeline-and-platform-evaluation-tool/vippet/pipelines/video-summarization-vlm.yaml
```diff
@@ -0,0 +1,58 @@
...
+      video/x-raw,format=RGB !
+      gvagenai
+        model-path=/models/output/openvino_models/cpu/INT4/google/gemma-3-4b-it
+        device=CPU
+        prompt="Summarize this video in one sentence."
+        generation-config="max_new_tokens=64"
+        frame-rate=1
+        chunk-size=4
+        metrics=false !
+      gvafpscounter starting-frame=100 !
```
p-zak: For a pipeline that runs the VLM on `chunk-size=4` frames at `frame-rate=1`, hitting frame 100 means the FPS counter only starts after a long warm-up. Is that the intent?

tools/visual-pipeline-and-platform-evaluation-tool/vippet/pipelines/video-summarization-vlm.yaml
```diff
@@ -0,0 +1,58 @@
...
+      gvametapublish method=file file-format=json-lines file-path=/videos/output/summary.jsonl !
```
p-zak: This pipeline does not use the existing metadata-publishing infrastructure that the rest of ViPPET relies on for `gvametapublish`. Instead it hardcodes a single, shared output path:

```
gvametapublish method=file file-format=json-lines file-path=/videos/output/summary.jsonl
```

For comparison, the existing flow (used since the gvametapublish/gvamotiondetect work) is:

- `Graph.inject_metadata_file_paths(metadata_dir)` (`graph.py` ~L1050) sets `method=file`, `file-format=json-lines`, and `file-path=<metadata_dir>/metadata_<node.id>.jsonl` on every `gvametapublish` node, **overwriting any existing values**.
- `pipeline_manager.build_pipeline_command` (~L627) calls it with `metadata_dir = make_output_dir(metadata_job_dir, pipeline_id)` - so the directory is unique per `(job_id, pipeline_id)` and the filename is unique per `gvametapublish` node id.
- Only injected when `metadata_mode == InternalMetadataMode.FILE` and `stream_index == 0`.
- The `/jobs/{job_id}/metadata/{pipeline_id}/{file}` REST endpoint and the corresponding SSE stream (`vippet/api/routes/jobs.py` L429+, L497+) read from exactly those files.

Because this YAML hardcodes `file-path=/videos/output/summary.jsonl`, several things break:

1. **Multi-stream within one pipeline (`streams > 1`)** - every stream writes to the **same** `summary.jsonl`. Multiple writers on a single file → interleaved/torn JSON lines and lost records; this is a real data-corruption issue, not just a UX one.
2. **Concurrent CPU/GPU/NPU variants (or repeated runs)** - same shared path, same race.
3. **Cross-job overwrites** - each new job overwrites the previous summary; every other pipeline gets per-`job_id/pipeline_id` isolation, this one does not.
4. **Invisible to the existing API.** The REST/SSE endpoints look under `metadata_job_dir/pipeline_id/`, but this pipeline writes to `/videos/output/`. So gvagenai output cannot be consumed by the existing UI / SSE clients without a separate code path.
5. **`metadata_mode = DISABLED` is ignored.** The file is always created regardless of the user's metadata-mode choice - the hardcoded `file-path` bypasses the `metadata_mode` flag entirely.
6. **`inject_metadata_file_paths` would normally save us** (it overwrites `file-path` if `metadata_mode == FILE`), but at `metadata_mode != FILE` (the default for many runs) the hardcoded path stays as-is, so points 1–5 still apply.

Suggested fix: drop the `file-path` (and optionally `method` / `file-format`) from the YAML and let the existing infrastructure inject them, exactly like every other built-in pipeline:

```yaml
gvagenai
  model-path=/models/output/openvino_models/cpu/INT4/google/gemma-3-4b-it
  device=CPU
  prompt="Summarize this video in one sentence."
  generation-config="max_new_tokens=64"
  frame-rate=1
  chunk-size=4
  metrics=false !
gvafpscounter starting-frame=100 !
gvametaconvert format=json !
gvametapublish method=file file-format=json-lines file-path=/dev/null !
fakesink sync=false async=false
```

With that change, gvagenai automatically gets:
- per-`(job_id, pipeline_id)` output directory,
- per-`gvametapublish`-node unique filename,
- correct `metadata_mode = DISABLED` behavior (no file written when the user disables it),
- multi-stream safety (each stream gets a stream-scoped path through the same machinery),
- visibility through the existing `/jobs/{job_id}/metadata/...` REST + SSE endpoints - no special-casing needed.

p-zak (changes requested):

tools/visual-pipeline-and-platform-evaluation-tool/vippet/api/api_schemas.py
Supriya-Krishnamurthi: Updated

tools/visual-pipeline-and-platform-evaluation-tool/vippet/api/api_schemas.py
Supriya-Krishnamurthi: Updated

tools/visual-pipeline-and-platform-evaluation-tool/vippet/graph.py
Supriya-Krishnamurthi: Updated

tools/visual-pipeline-and-platform-evaluation-tool/vippet/pipelines/video-summarization-vlm.yaml
Supriya-Krishnamurthi: At 30 fps, frame-rate=1 keeps one frame/sec — enough to catch scene change without flooding the VLM. chunk-size=4 gives 4 s of temporal context per VLM call, the minimum to distinguish dynamic states ("people walking" vs "stopped"). Result: one summary every 4 s — a readable UI flow.

starting-frame=100 was chosen assuming long surveillance-style videos may be used for benchmarking, where skipping the first ~100 frames avoids skewing gvafpscounter's average with model-load and first-token latency and other pipeline ramp-ups. You're right it's too aggressive for short clips. Lowered to 10 as a safer default for mixed input lengths.

tools/visual-pipeline-and-platform-evaluation-tool/vippet/pipelines/video-summarization-vlm.yaml
Supriya-Krishnamurthi: updated `gvametapublish file-path=/dev/null` so the injector overwrites it for stream 0 / metadata_mode=file, and discards cleanly for streams ≥ 1 as they are streams of same pipleine and outputs are near duplicates.

For this pipeline, **removing gvametaconvert** would make UI less noisy:

gvametaconvert is useful when multiple inference elements (e.g. gvadetect + gvaclassify) need to follow one schema (`{"objects":[…], "resolution":…, "timestamp":…}`). This pipeline has only gvagenai, which already emits a JSON-friendly meta that gvametapublish writes cleanly as `{"result", "confidence", "timestamp"}`

The issue: gvametaconvert re-stamps the last seen result onto every buffer that flows through, so results in duplicated summaries across many frames.

Example : Input video is 10 s @ 30 fps, frame-rate=1, chunk-size=4:

300 total frames; first chunk consumes 4 s → 120 frames withheld → 180 reach the publisher.
VLM produces only [10/4]≈2 summaries

Without convert: 2 lines in ouput file. With convert: ~180 lines, only 2 unique summaries → **99% duplication**.
UI consumes this output file so it makes output look duplicated with ~180 noisy results containing only 2 unique summary). Removing it gives a 1:1 mapping between summaries and metadata records. Can be reintroduced if this pipeline is later combined with frame-rate inference elements.

p-zak: Good catch with `gvametaconvert`!

p-zak (approved):

tools/visual-pipeline-and-platform-evaluation-tool/shared/models/supported_models.yaml
```diff
@@ -280,3 +280,15 @@
...
+  display_name: Gemma 3
+  source: huggingface
+  type: genai
+  precisions:
+    - precision: INT4
+      model_path: openvino_models/cpu/INT4/google/gemma-3-4b-it/
+      model_proc: ""
+  extra_model_procs: []
+  unsupported_devices: ""
+  default: false
```
dlemiech, line 294: missing EOF

dlemiech (approved):

tools/visual-pipeline-and-platform-evaluation-tool/shared/models/supported_models.yaml
p-zak: It works, but I will fix it in my PR (related to model-download service)

PR #2181, opened 2026-05-07
Hardened Segment Config Processing in MME and Bug Fixes.

sun1lach:
### Description

 - Added `segment_config` processing in `decoder.py`
 - Priority processing: frame_indices > extraction_fps > num_frames from cURL.
 - Removed `decord` dependency.
 - Update `EMBEDDING_USE_OV` defaults in `setup.sh`.

Fixes:
91454

No

### How Has This Been Tested?
Dev Environment

### Checklist:


microservices/multimodal-embedding-serving/src/wrapper.py
```diff
@@ -363,16 +352,6 @@ async def get_video_embedding_from_frames_manifest(self, manifest_path: str) ->
...
                 }
                 
-                # Extract specified frames from video
-                extracted_frames = extract_video_frames(
-                    safe_manifest_video_path, segment_config
-                )
-                
-                if not extracted_frames:
-                    raise ValueError(
-                        f"No frames could be extracted from video: {sanitize_for_log(safe_manifest_video_path)}"
```
yogeshmpandey, line 373: Is this check not needed now ?

microservices/multimodal-embedding-serving/src/utils/decoder.py
```diff
@@ -288,6 +289,96 @@ def effective_workers(self) -> int:
...
+    Args:
+        total_frames: Total number of frames in the video
+        fps: Frames per second of the video
+        stream_config: Video frame configuration with segment parameters
+
+    Returns:
+        Set of frame indices to extract
+    """
+    # Calculate segment boundaries
+    start_idx = int(fps * stream_config.start_offset_sec)
```
yogeshmpandey, line 314: Not sure but if fps is float then will we not miss some frames ?

microservices/multimodal-embedding-serving/src/wrapper.py
sun1lach: Previous implementation was a 2-step process - `extract_video_frames` and use that frames for embedding generation.

In L419, this logic is already handled and delegated to `self.get_video_sampled_embeddings(...)`. So the additional check for the manifest flow is not needed.

microservices/multimodal-embedding-serving/src/utils/decoder.py
sun1lach: This logic is same as what we had for `decord` based implementation.
Retaining the same steps.
Also, the `start_idx` has to be int to perform list/np.ndarray range query or it's use in `np.linspace(...)`.

yogeshmpandey (approved):

14pankaj (approved):

PR #2206, opened 2026-05-12
[MME] PyTorch and Model Variant support.

sun1lach:
### Description

 - PyTorch fallback support for non-OV scenario.
 - Fallback for all supported model variants.
 - Handle preprocess metadata support in blip2 transformers.
 - Validated on CPU (PyTorch/OpenVINO) & GPU (OpenVINO) for all supported models.

Fixes 
91454
91456

### How Has This Been Tested?

Dev local - ARL 255H

### Checklist:


bharagha (approved):

microservices/multimodal-embedding-serving/src/models/handlers/blip2_handler.py
```diff
@@ -208,39 +208,41 @@ def encode_text(self, texts: Union[str, List[str]]) -> torch.Tensor:
         text_features = F.normalize(text_features, dim=-1)
         return text_features
     
-    def encode_image(self, images: Union[Image.Image, List[Image.Image]], metrics_out: bool = False) -> Union[Dict[str, Any], np.ndarray]:
-        """Encode images using SigLIP image encoder."""
+    def encode_image(self, images: Union[Image.Image, List[Image.Image]], metrics_out: bool = False) -> Union[Dict[str, Any], torch.Tensor]:
+        """Encode images using BLIP-2 image encoder."""
```
yogeshmpandey, line 212: Any reason for this change. I think SigLIP is more efficient right? is it for accuracy

microservices/multimodal-embedding-serving/setup.sh
```diff
@@ -46,6 +46,7 @@ export VIDEO_FRAME_LOG_LEVEL=${VIDEO_FRAME_LOG_LEVEL:-INFO}
 # If EMBEDDING_DEVICE is GPU, set EMBEDDING_USE_OV to true
 if [ "$EMBEDDING_DEVICE" = "GPU" ]; then
     export EMBEDDING_USE_OV=true
+    export OV_PERFORMANCE_MODE=THROUGHPUT
```
yogeshmpandey, line 49: export OV_PERFORMANCE_MODE=${OV_PERFORMANCE_MODE:-THROUGHPUT} ?

microservices/multimodal-embedding-serving/src/models/handlers/blip2_transformers_handler.py
```diff
@@ -440,28 +440,28 @@ def encode_image(self, images: Union[Image.Image, List[Image.Image]], metrics_ou
...
-                    "embeddings": embeddings,
+                    "embeddings": image_features,
                     "inference_time_s": infer_end - infer_start,
-                    "processed_images": total_images
+                    "processed_images": total_images,
                 }
-            image_features = torch.from_numpy(embeddings)
         else:
             # Use PyTorch model (already includes projection + normalization)
             if isinstance(images, torch.Tensor):
```
yogeshmpandey, line 462: Is this not removed in base ? is torch.Tensor needed ?

microservices/multimodal-embedding-serving/src/models/handlers/cn_clip_handler.py
```diff
@@ -230,81 +252,103 @@ def _encode_text_openvino(self, texts: List[str]) -> torch.Tensor:
...
+
         if self.use_openvino:
-            return self.ov_image_encoder.outputs[0].shape[1]
+            if self.async_infer is None:
+                raise RuntimeError("OpenVINO async inference not initialized. Call load_model() first.")
+
+            dummy_image = Image.fromarray(np.random.randint(0, 255, (image_size, image_size, 3), dtype=np.uint8))
+            dummy_np = np.array(self.preprocess(dummy_image))[np.newaxis]  # [1, C, H, W]
+            result = self.async_infer.infer(dummy_np)
+            self._embedding_dim = int(result.shape[-1])
```
yogeshmpandey, line 319: Not able to understand this snippet. Do we need to run the inference call here ? to get the dimensions ?

microservices/multimodal-embedding-serving/src/models/handlers/blip2_handler.py
yogeshmpandey: my bad , I see thats doc change

microservices/multimodal-embedding-serving/setup.sh
sun1lach: It's getting set in L32.

In L49, we are enforcing throughput because latency is better on CPU and throughput is better on GPU.

microservices/multimodal-embedding-serving/src/models/handlers/blip2_transformers_handler.py
sun1lach: Re-instated `torch.Tensor` in the base and other function signature as it's a zero-copy view over numpy array.
Other MME Rest API (base64, URL, etc) in wrapper.py performs norm on returned tensor. If i return `np.ndarray`, it doesn't have `np_array.norm` and it fails.

To keep it all consistent and backward compatible, retaining torch.Tensor across.

microservices/multimodal-embedding-serving/src/models/handlers/cn_clip_handler.py
sun1lach: In other existing clip-based handlers, `get_embedding_dim(..)` does this operation (It does inference on dummy image, checks output dimension and returns that).

In `cn_clip_handler.py` it was missing. So, I copied it here.

If you ask me, there are better way to get model's output dimension. 
Just retaining what we had before. I'll revisit this later to update `get_embedding_dim(..)` function logic in a better way.

yogeshmpandey: ok

microservices/multimodal-embedding-serving/src/models/handlers/blip2_transformers_handler.py
yogeshmpandey: ok

yogeshmpandey (approved):

PR #2250, opened 2026-05-18
[ViPPET] Update metrics charts

KaliszWiktoria:
Updated metrics dashboard:
- Divided dashboard into two columns of charts
- Made charts windows the same length
- Provided switch for Metadata Json, where disabling metadata provides it from being sent from backend
- When switch is disabled, tabs for output video and metadata won't show when not necessary

**PREVIEW:**

https://github.com/user-attachments/assets/b34b51bc-d2a7-4ecf-9be2-d614deb0aa03


p-zak (approved):

staszczuk (approved):
Beautiful

PR #2316, opened 2026-05-28
chatqna-core: Enforce version-pinned defaults for backend and UI images

hteeyeoh:
### Description

As chatqna-core and chatqna modular shared the same image name. Tag 'latest' has been assigned for chatqna modular.Avoid fallback to `latest` when env vars are unset to ensure reproducible deployments.

Fixes # (issue)

N/A

### How Has This Been Tested?

### Checklist:

### DIFF
--- a/sample-applications/chat-question-and-answer-core/docker/compose.yaml
+++ b/sample-applications/chat-question-and-answer-core/docker/compose.yaml
@@ -7,7 +7,6 @@ x-nginx-common: &nginx-common
     - ../nginx_config/nginx.conf:/etc/nginx/nginx.conf:ro
 
 x-chatqna-core-common: &chatqna-core-common
-  image: ${REGISTRY:-}chatqna:${BACKEND_TAG:-latest}
   build:
     context: ../
     dockerfile: docker/Dockerfile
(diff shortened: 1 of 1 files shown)


sample-applications/chat-question-and-answer-core/docker/compose.yaml
```diff
@@ -53,12 +52,14 @@ services:
     <<: *chatqna-core-common
     profiles:
```
bharagha, line 53: @hteeyeoh : Changes like this should only be for release branch. Not mainline. Please move the changes to release branch.

bharagha (changes requested):

sample-applications/chat-question-and-answer-core/docker/compose.yaml
hteeyeoh: ok. The issue here is that now the tag `latest` is being used by chatqna-modular as both modular and core shared same image name. There will be risk where docker might pull this chatqna:latest from dockerhub where it actually is a chatqna-modular image.  If it is ok to keep the tag as latest in main. I can change.

hteeyeoh:
Changes not required for main branch

PR #2472, opened 2026-06-24
VSS - Fix/search ms bugs

bhardwaj-nakul:
### Description

 fix(vss): align search filters and storage-key usage across manager/search-ms
 
 - accept camelCase/snake_case query fields in search-ms (queryId/timeFilter aliases)
 - fix tag subset filtering semantics in search-ms query results
 - forward tags in pipeline-manager /search/query requests
 - support absolute time windows (start/end ISO) in search-state normalization
 - allow optional tags in /videos/search-embeddings/:videoId and merge/persist them
 - send tags from UI for existing-video embeddings and refresh videos/tags after success
 - make datastore object keys deterministic (<videoId>/source.<ext>) and encode object URL paths
 - use storage object filename (from video url) for audio transcription and embedding payloads
 - update related unit tests/specs for datastore, search, video, and audio queue flows

Fixes # (issue)

None

### How Has This Been Tested?

Tested in dev env

### Checklist:


sample-applications/video-search-and-summarization/search-ms/server.py
```diff
@@ -58,35 +65,25 @@ def _build_explicit_time_filter(time_filter: Optional[TimeRange], property_name:
...
-    if not effective_time_filter and not tag_filter:
+    if not effective_time_filter:
         return None, None
 
-    if effective_time_filter and tag_filter:
-        merged_filter = {**effective_time_filter, **tag_filter}
-        return merged_filter, effective_time_filter
-
-    return effective_time_filter or tag_filter, effective_time_filter
+    return effective_time_filter, effective_time_filter
```
yogeshmpandey: is there a typo here ? Not able to understand this

bhardwaj-nakul: Addressed. I refactored this helper now to return a single `Optional[dict]` filter value (removed the redundant tuple return) and updated the caller accordingly, so the intent is explicit and there is no duplicate return shape anymore. I also kept concise inline comments in the new storage-filename handling paths to clarify why object-key-derived filenames are used downstream.

yogeshmpandey (approved):

krish918 (approved):

PR #2495, opened 2026-07-02
feat: Add feature to export platform and device capabilities

ArokiEdgard:
### Description

Please include a summary of the changes and the related issue. List any dependencies that are required for this change.

Fixes # (issue)

### Any Newly Introduced Dependencies

This change does not add any new Python package dependencies. It does introduce two runtime system packages in microservices/metrics-manager/Dockerfile,
**dmidecode:** It is released under the GNU General Public License (GPL). Used by the capabilities feature to read SMBIOS/DMI data and enrich platform memory information, especially memory technology/type such as DDR4 or DDR5.

pciutils (lspci): GNU General Public License, version 2 or any later version. Used by the capabilities feature for best-effort PCI device branding, especially readable GPU model/vendor naming in the capabilities JSON

### How Has This Been Tested?
Test environment:
Linux host
Docker Compose deployment from microservices/metrics-manager
Host visibility provided through existing runtime mounts and settings:
/sys:/sys:ro
/run:/run:ro
pid: host
dri device mapping for GPU visibility
privileged: true for NPU and hardware inventory scenarios

1. Recreated the service with Docker Compose
cd edge-ai-libraries/microservices/metrics-manager
docker compose up -d --force-recreate metrics-manager
2. Verified container health
curl -s http://127.0.0.1:9090/health | jq 
3. Verified capabilities API, minimal profile
curl -s "http://127.0.0.1:9090/api/v1/capabilities?profile=minimal" | jq
4. Verified capabilities API, expanded profile
curl -s "http://127.0.0.1:9090/api/v1/capabilities?profile=expanded" | jq
5. Verified representative capability fields from the host
curl -s "http://127.0.0.1:9090/api/v1/capabilities?profile=minimal" | jq '.profile, .platform.hostname, .platform.device_summary'

Observed behavior during validation:

- The capabilities API returned valid JSON from the host on port 9090.
- The minimal profile returned platform hostname and device summary.
- The expanded profile returned enriched platform/device inventory.

### Checklist:

- [x] I agree to use the APACHE-2.0 license for my code changes.
- [x] I have not introduced any 3rd party components incompatible with APACHE-2.0. 
- [x] I have not included any company confidential information, trade secret, password or security token. 
- [x] I have performed a self-review of my code.

PR #2551, opened 2026-07-14
Automated benchmarking of VIPPET pipelines

nathsudi:
### Description

Automated benchmarking of VIPPET pipelines — CPU, GPU (Xe), and NPU. Collects hardware KPIs per job and produces JSON, CSV, and HTML bar-chart reports.

Fixes # (issue)

### How Has This Been Tested?

### Checklist:


tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/scripts/benchmark.py
```diff
@@ -0,0 +1,324 @@
+#!/usr/bin/env python3
```
kkepka: please remember to add copyright and license headers, like:

```
# Copyright (C) 2026 Intel Corporation
# SPDX-License-Identifier: Apache-2.0
```

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/src/hw_monitor.py
```diff
@@ -0,0 +1,189 @@
...
+
+Samples CPU, GPU (Xe), and NPU KPIs in a background thread during benchmark runs.
+
+Source: VIPPET metrics-manager JSON API (CPU, GPU, NPU, memory, temperature, power).
+"""
+
+import logging
+import threading
+import time
+from typing import Any, Dict, List, Optional
```
kkepka: please avoid legacy imports from typing, unless necessary
for reference see: https://docs.openedgeplatform.intel.com/dev/edge-ai-libraries/visual-pipeline-and-platform-evaluation-tool/developer-guide/contributing/backend.html#python-typing-python-3-12

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/src/vippet_client.py
```diff
@@ -0,0 +1,297 @@
...
+VIPPET REST API Client
+
+Provides interface to interact with VIPPET REST API endpoints.
+"""
+
+import time
+import logging
+from typing import Dict, List, Any, Optional
+
+import requests
```
kkepka: in functional tests `httpx` is used for generating client requests,
please possibly switch to that one, to avoid adding a new dependency

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/scripts/generate_report.py
```diff
@@ -0,0 +1,953 @@
...
+    runs = load_results(paths)
+    html = generate_html(runs)
+
+    if args.output:
+        out = Path(args.output)
+    else:
+        out = paths[0].with_suffix(".html")
+
+    out.write_text(html)
+    print(f"Report saved: {out}")
```
kkepka: please rather use logging module, instead of `print`'s

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/config/default.yaml
```diff
@@ -0,0 +1,76 @@
+# VIPPET Benchmark Suite - Default Configuration
+
+vippet:
+  # VIPPET API base URL
+  base_url: "http://localhost:7860/api/v1"
```
kkepka: vippet's standard convention is to use port 80 by default

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/src/vippet_client.py
```diff
@@ -0,0 +1,289 @@
...
+
+    def __init__(self, base_url: str, timeout: int = 600):
+        """
+        Initialize VIPPET client.
+
+        Args:
+            base_url: VIPPET API base URL (e.g., http://localhost:7860/api/v1)
+            timeout: Default timeout for API requests in seconds
+        """
+        self.base_url = base_url.rstrip("/")
```
kkepka: typically (e.g. in [functional tests](https://github.com/open-edge-platform/edge-ai-libraries/blob/9bb73d1ef232bd17d1fad62db5afc9c403ee78cd/tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/functional/helpers/config.py#L6)) `VIPPET_BASE_URL` is read, could you please consider adding support for setting it within tests itself? (as right now it's set via `--vippet-url` in run.sh)

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/run.sh
```diff
@@ -0,0 +1,89 @@
...
+
+# ── check Python deps ─────────────────────────────────────────────────────────
+info "Checking Python dependencies..."
+if ! python3 -c "import requests, yaml" 2>/dev/null; then
+    warn "Installing missing dependencies..."
+    pip3 install -q -r "$SUITE_DIR/requirements.txt"
+fi
+
+# ── check VIPPET is reachable ─────────────────────────────────────────────────
+VIPPET_URL="${VIPPET_URL:-http://localhost:7860/api/v1}"
```
kkepka: please rename to `VIPPET_BASE_URL` for consistency across tests

kkepka (commented):
Thank you @nathsudi for this contribution, appreciated.

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/config/default.yaml
nathsudi:
```
1b3afa9c16df   intel/vippet-ui:2026.1.0-20260512-weekly                "/docker-entrypoint.…"   4 hours ago   Up 4 hours             0.0.0.0:80->80/tcp, [::]:80->80/tcp                                                                                                     ui
a6b2f680d238   intel/vippet-app:2026.1.0-20260512-weekly               "./entrypoint.sh"        4 hours ago   Up 4 hours (healthy)   0.0.0.0:7860->7860/tcp, [::]:7860->7860/tcp
```

UI is running in port 80, but backend is running in 7860 when I deployed using handheld-multimodal-app package

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/scripts/benchmark.py
nathsudi: Added in all files

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/src/hw_monitor.py
nathsudi: Replaced legacy typing imports (Dict, List, Optional) with Python 3.12 built-in generics (dict, list, X | None)

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/src/vippet_client.py
nathsudi: Switched from requests to httpx

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/scripts/generate_report.py
nathsudi: Replaced print() with logging module

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/src/vippet_client.py
nathsudi: Added VIPPET_BASE_URL env var support in client (os.environ.get("VIPPET_BASE_URL", base_url))

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/run.sh
nathsudi: Renamed VIPPET_URL to VIPPET_BASE_URL

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/config/default.yaml
kkepka: Port 7860 is the internal backend port, but port 80 (via the nginx reverse proxy in the ui container) is the documented public entry point. The nginx config at `ui/nginx.conf` routes `location /api/` -> `proxy_pass http://vippet:7860`, so `http://localhost/api/v1/*` reaches the same backend.

Using port 80 as the default is preferred because:
- Consistency - functional tests already default to `http://localhost/api/v1`
- Production parity - real deployments and the Helm chart expose port 80; tests should exercise the same path clients use
- Metrics proxy - localhost:9090 is also proxied through nginx at `/metrics/`, so using port 80 for both avoids coupling to internal port assignments

The benchmark suite should default to http://localhost/api/v1 (and http://localhost/metrics/stream for metrics). Port 7860 still works as an override via `--vippet-url` or `VIPPET_BASE_URL` for setups without the UI container, but shouldn't be the default.

nathsudi: Thanks got it. Replacing to use 80 in the next restructuring commit.

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/README.md
```diff
@@ -0,0 +1,66 @@
...
+## Prerequisites
+
+- VIPPET running at `http://localhost:7860` with models downloaded
+- Python dev venv set up (`make` from project root creates `.venv` automatically)
+
+## Usage
+
+```bash
+make test-performance                    # default: CPU+GPU+NPU, 1 and 3 streams
+make test-performance PERF_CONFIG=quick  # CPU+GPU only, 1 and 3 streams
```
kkepka: would you be kind to verify documentation matching with code please?
seems there is some divergence after reworks, e.g. current `quick` config aims for single stream only

kkepka (commented):
i did a verify run, where it got executed on my laptop, pointing with env vars to remote nuc running vippet

the whole `system` card contains information from where scripts got executed, not remote vippet target pointed with env vars, not sure if thats intended (i.e. is it assumed it will always be run at same machine?)

Image

it was also missing `GPU Util` stats in all scenarios

Image

nathsudi:
> i did a verify run, where it got executed on my laptop, pointing with env vars to remote nuc running vippet
> 
> the whole `system` card contains information from where scripts got executed, not remote vippet target pointed with env vars, not sure if thats intended (i.e. is it assumed it will always be run at same machine?)
> 
> Image
> it was also missing `GPU Util` stats in all scenarios
> 
> Image

Yes the sysinfo returning the runner sysinfo is intentional matching m9. 

For blank GPU util, 
 In `hw_monitor.py:36-39`, the SSE fetch reads lines until it finds the first data: line, then immediately breaks and closes the connection. If the metrics-manager hasn't
  emitted a data: event yet (e.g., it sends heartbeat/comment lines first, or its internal polling interval hasn't ticked), the 5-second timeout expires with `raw = ""`
  and the function returns {}. Empty samples are discarded at line 131, leading to `sample_count: 0`.

  Proposed Fix: Retry with backoff

  The simplest and most effective approach is to retry the fetch within _fetch_metrics_manager if the first attempt returns empty. Something like:

```
  # Try up to 3 times with a short delay between attempts
  for attempt in range(3):
      resp = requests.get(url, timeout=5, stream=True)
      resp.raise_for_status()
      raw = ""
      for line in resp.iter_lines(decode_unicode=True):
          if isinstance(line, str) and line.startswith("data: "):
              raw = line[6:]
              break
      resp.close()
      if raw:
          break
      time.sleep(1)
```

  This handles the case where the SSE endpoint is slow to emit its first event without changing the overall architecture. The 1-second sleep between retries is short enough that it won't significantly delay sampling, but gives the metrics-manager time to produce data. **Please let me know if you have some better idea for this.**

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/README.md
nathsudi: `quick.yaml` was changed (in commit c1638107) to only have stream_counts: [1]. The README was never updated. Thanks for noticing. Will check if there is any other doc misalignment and fix in next commit.

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/perf_helpers/hw_monitor.py
```diff
@@ -0,0 +1,230 @@
...
+        # GPU (Xe) — per-engine utilization
+        for engine, label in [
+            ("rcs", "gpu_render_util_pct"),
+            ("vcs", "gpu_video_util_pct"),
+            ("vecs", "gpu_enhance_util_pct"),
+            ("ccs", "gpu_compute_util_pct"),
+        ]:
+            val = mm.get(f"gpu_engine_usage_usage__{engine}")
+            if val is not None:
+                sample[label] = round(val, 2)
```
kkepka: continuation of:
- https://github.com/open-edge-platform/edge-ai-libraries/pull/2551#pullrequestreview-4723517506
- https://github.com/open-edge-platform/edge-ai-libraries/pull/2551#issuecomment-5015430468

while retry capability is rather a good thing to have, it's not actually the cause of missing gpu util metrics on my end.
metrics-manager (via qmassa) emits engine names as long-form strings ("render", "video", "video-enhance", "compute", "copy") for both i915 and Xe drivers. short engine-class codes (rcs/vcs/vecs/ccs/bcs) are kept as fallback aliases in case an older qmassa version emits those instead.
please see suggestion below (which works fine on my setup)

```suggestion
        # GPU (Xe/i915) — per-engine utilization
        for label, aliases in [
            ("gpu_render_util_pct", ("render", "rcs")),
            ("gpu_video_util_pct", ("video", "vcs")),
            ("gpu_enhance_util_pct", ("video-enhance", "vecs")),
            ("gpu_compute_util_pct", ("compute", "ccs")),
            ("gpu_copy_util_pct", ("copy", "bcs")),
        ]:
            for engine in aliases:
                val = mm.get(f"gpu_engine_usage_usage__{engine}")
                if val is not None:
                    sample[label] = round(val, 2)
                    break
```

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/perf_helpers/reporters.py
```diff
@@ -0,0 +1,805 @@
...
+                "cpu_util": _safe(hw.get("cpu_util_pct_avg"), 1),
+                "cpu_temp": _safe(hw.get("cpu_temperature_avg"), 1),
+                "cpu_temp_max": _safe(hw.get("cpu_temperature_max"), 1),
+                "error": tc.get("error"),
+            }
+        )
+    return rows
+
+
+def _build_sysinfo_cards(system_info: dict[str, Any]) -> str:
```
kkepka, line 437: continuation of:
- https://github.com/open-edge-platform/edge-ai-libraries/pull/2551#pullrequestreview-4723517506
- https://github.com/open-edge-platform/edge-ai-libraries/pull/2551#issuecomment-5015430468

collecting system info from local machine and presenting on report with performance data for remote target might be confusing for readers, would you agree?
it might be better to avoid such confusion by creating report with remote/target system information in report card

vippet exposes `/devices` endpoint, from where we can fetch information about CPU, GPU and NPU already

recently there was also added #2495 , what might be another source of more detailed system information

nathsudi: `collecting system info from local machine and presenting on report with performance data for remote target might be confusing for readers, would you agree?`
I agree on this. If vippet exposes the device details then it would be better to extract from it and display.

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/config/quick.yaml
```diff
@@ -0,0 +1,36 @@
...
+vippet:
+  base_url: "http://localhost/api/v1"
+  timeout: 300
+  poll_interval: 2
+  max_job_duration: 300
+
+benchmark:
+  pipelines: "*"
+
+  variants:
```
kkepka, line 13: could you please verify if vars from config are properly used during test runs?

i.e. if we would change in `quick.yml`:
`pipelines` to `["license-plate-recognition", "object-detection"]`
`variants` to just `gpu`,
`stream_counts` to e.g. `3`
and then run `make test-performance PERF_CONFIG=quick`

then only last is being applied, i.e. first test running is `test_pipeline_performance: Running performance benchmark: pipeline='Age & Gender Recognition' variant=CPU streams=3`, which is a bit unexpected, would you agree?

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/perf_helpers/hw_monitor.py
```diff
@@ -0,0 +1,233 @@
...
+Source: VIPPET metrics-manager JSON API (CPU, GPU, NPU, memory, temperature, power).
+"""
+
+import json
+import logging
+import threading
+import time
+from typing import Any
+
+import requests
```
kkepka: seems `requests` are re-introduced ([ref](https://github.com/open-edge-platform/edge-ai-libraries/pull/2551#discussion_r3576467998))

nathsudi: not re-introduced, this one escaped from earlier cleanup.

tools/visual-pipeline-and-platform-evaluation-tool/vippet/tests/performance/config/quick.yaml
nathsudi: all vars in yaml was not wired, added the missing config load.

kkepka (approved):
Scope summary - automated performance benchmark framework that:
- Discovers all runnable pipeline/variant combinations from the live VIPPET API
- Runs each with configurable stream counts and collects hardware KPIs in a background thread
- Produces multi-format reports (JSON/CSV/HTML) with per-pipeline FPS bar charts and utilization data

Lets treat this as good incremental step towards more robust performance testing, as early-preview/experimental feature.

Thank you for this contribution!

grey-eminence (approved):

PR #2584, opened 2026-07-21
GENAI - Model-download-Makefile added for deployments and Tests

saikiransayabugari:
## Summary

Adds a standardized `Makefile` (plus supporting `.env.example`) for the `model-download` microservice, providing a single entry point for build/deploy/test/lint and security-scan workflows — consistent with the pattern used by other components (e.g. `video-search-and-summarization`).

## Changes

- **`microservices/model-download/Makefile`** (new): thin wrapper around existing scripts/tooling with targets:
  - `build` / `deploy` / `deploy-down` — wrap `scripts/run_service.sh`
  - `test` - runs `uv sync --extra dev --extra geti` + `pytest` with coverage (term, HTML)
  - `lint` - `pylint` on `src/`
  - `shellcheck` - ShellCheck (JSON report) on shell scripts
  - `pylint` - Pylint JSON security report
  - `trivy-scan` (`-fs`, `-image`, `-config`) — filesystem, image, and Dockerfile scans (JSON/SPDX/HTML)
  - `clamav-scan` - ClamAV antivirus scan
  - `bandit-scan` - Bandit security scan (txt/json/html), run in an isolated venv
  - `gitleaks-scan` - Gitleaks secrets scan, **auto-installs** the correct OS/arch `gitleaks` binary into `.tools/` if not already on `PATH` (no sudo, no global PATH changes)
  - `codeql-scan` - CodeQL Python analysis (SARIF + CSV), **auto-installs** the CodeQL CLI into `.tools/codeql/` if missing
  - `clean` - removes caches, coverage artifacts, `.tools/`, `bandit-venv/`, `.pylintrc`, and `.venv`
  - `help` - lists all available targets
  - Auto-creates `.env` from `.env.example` on first `build`/`deploy`/`deploy-down` if not already present
- **`microservices/model-download/.env.example`** (new): template environment file for local configuration (registry/tag, secrets placeholders, etc.), copied to `.env` automatically on first run.
- **`microservices/model-download/scripts/run_service.sh`**: minor fix — `docker compose up` → `docker compose up -d` so `make deploy` doesn't block the terminal.

## Testing

- `make gitleaks-scan` - verified end-to-end on a machine without `gitleaks` preinstalled: auto-downloads v8.30.1, scans, writes report to `security-results/`.
- `make codeql-scan` - verified end-to-end on a machine without `codeql` preinstalled: auto-downloads CLI v2.26.1, creates DB, produces SARIF + CSV reports.
- `make test` - fixed a bug where `tests/unit/test_geti_plugin.py` (which imports `geti_sdk`, gated behind the `geti` extra, not `dev`) aborted the *entire* pytest collection (0/253 tests actually ran, silently masked by `|| true`). Now installs `--extra geti` as well; full suite of 256 tests collects and runs (230 passed, 26 pre-existing failures unrelated to this change - tracked separately).
- `make clean` - verified removes all residue from `test`/`lint`/scan targets (`.venv`, `security-results/`, `.tools/`, `bandit-venv/`, coverage artifacts).

## Notes / Follow-ups

- CI (`GENAI-model-download.yaml`) currently reimplements each scan inline instead of calling these Makefile targets, and has no `codeql-scan` job — aligning CI with the Makefile (installing tools then calling `make <target>`, as `GENAI-VideoSearchAndSummarization.yaml` does) is planned as a follow-up, not part of this PR.


madhuri-rai07 (dismissed):

microservices/model-download/Makefile
```diff
@@ -0,0 +1,346 @@
...
+.PHONY: help build deploy deploy-down test lint clean \
+        shellcheck pylint \
+        trivy-scan trivy-scan-fs trivy-scan-image trivy-scan-config \
+        clamav-scan bandit-scan gitleaks-scan codeql-scan
+
+help:
+	@echo "Model Download - Makefile Commands"
+	@echo "==================================="
+	@echo "build   - Build the Docker image (scripts/run_service.sh --build)"
+	@echo "deploy      - Start the service (scripts/run_service.sh up)"
```
Supriya-Krishnamurthi, line 64: We also pass plugins to be activated, and model-path while bringing up the service. How do we handle that here?

saikiransayabugari: The following ways can be used: 
1. You can use the .env.example file which has all the default values and if you need any new values, Just change the default values in the file.
2. Variables can be passed during the makefile runtime. For example
make deploy MODEL_PATH="$PWD" ENABLED_PLUGINS="ollama"

Supriya-Krishnamurthi: Okay, one example we can add : make deploy MODEL_PATH="$PWD" ENABLED_PLUGINS="ollama"

saikiransayabugari: okay sure.

saikiransayabugari: added those changes.

Supriya-Krishnamurthi (approved):

14pankaj (approved):

PR #2812, opened 2026-08-13
[ViPPET] Fix genai model falsely reported as installed after incomplete download

KaliszWiktoria:
## Problem

GenAI models (e.g. Gemma 3) were reported as **INSTALLED** in the UI even when the download had failed silently — for example because `HF_TOKEN` was not set for a gated HuggingFace model.

Running the pipeline with a falsely-installed model fails with:
```
Pipeline execution failed: Failed to create OpenVINO backend: ir: Could not open
the file: ".../openvino_language_model.xml"
```

### 1. `models.py` — `SupportedModel.exists_on_disk()`

**Before:** Returned `True` if the genai model directory existed.  
**After:** Also requires `openvino_language_model.xml` to be present directly in that directory. This file is the main language model artefact written by the OpenVINO GenAI download pipeline; its absence proves the download did not complete.

### 2. `model_manager.py` — `_load_registry()`

**Before:** Pruned stale `installed_models.json` entries using `os.path.exists(model_path)`, which returned `True` for a partial genai directory, keeping the stale `INSTALLED` registry entry alive across restarts.  
**After:** Introduces `_precision_is_complete(category, model_path)` helper that applies the same sentinel-file check for `genai` category models. The stale entry is now pruned on startup.

### 3. `model_manager.py` — `_finalize_success()`

**Before:** Unconditionally registered the model as `INSTALLED` and wrote it to `installed_models.json` whenever the model-download microservice reported `"completed"` — even when that "success" was silent (files not actually written).  
**After:** Calls `exists_on_disk()` on each YAML entry before registering. If the sentinel file is absent despite the service reporting success, the job is **reclassified as `FAILED`** with a human-readable message pointing to `HF_TOKEN` as the likely cause.

The same `_precision_is_complete()` helper is also applied in `_fail_job()` when cleaning up the registry after a re-download failure.

Tests have been updated

image


tools/visual-pipeline-and-platform-evaluation-tool/vippet/managers/model_manager.py
```diff
@@ -1326,6 +1365,25 @@ def _finalize_success(
...
+                "model-download reported success for '%s' (job %s) but the "
+                "expected model files are not present on disk — reclassifying "
+                "as FAILED.  If this is a gated model, ensure HF_TOKEN is set "
+                "correctly before retrying.",
+                model_name,
+                job_id,
+            )
+            self._fail_job(
+                job_id,
+                f"Model files not found after download reported success. "
```
dlemiech: I would rephrase a little bit:

Model <name> was not successfully installed. Check your Hugging Face access token and accept model license if needed.

dlemiech (dismissed):

tools/visual-pipeline-and-platform-evaluation-tool/vippet/managers/model_manager.py
KaliszWiktoria: Addressed in: https://github.com/open-edge-platform/edge-ai-libraries/pull/2812/changes/49c167b8f905b8c44aef66c623d4bdd8dc6cb070

dlemiech (dismissed):

tools/visual-pipeline-and-platform-evaluation-tool/vippet/managers/model_manager.py
```diff
@@ -1326,6 +1365,23 @@ def _finalize_success(
...
+                "expected model files are not present on disk — reclassifying "
+                "as FAILED.  If this is a gated model, ensure HF_TOKEN is set "
+                "correctly before retrying.",
+                model_name,
+                job_id,
+            )
+            self._fail_job(
+                job_id,
+                f"Model was not successfully installed. "
+                f"Check your Hugging Face access token and accept model license if needed. ",
```
kkepka: these are not formatted strings, please remove `f` prefix
(and please run `make ruff` and `make pyright` to check for other similar checks)

tools/visual-pipeline-and-platform-evaluation-tool/vippet/managers/model_manager.py
```diff
@@ -98,6 +98,29 @@
 # Upload streaming chunk size.
 UPLOAD_CHUNK_SIZE: int = 8 * 1024 * 1024  # 8 MiB
 
+# Main language model file that every OpenVINO GenAI model must contain.
+# Its presence proves the download completed successfully; a directory that
+# only contains early-stage artefacts (e.g. .cache/ or graph.pbtxt written
+# before the weights are fetched) will not contain this file.
+_GENAI_SENTINEL_FILE: str = "openvino_language_model.xml"
```
kkepka: is `openvino_language_model.xml` guaranteed for all currently supported and near-term genai models across our supported_models (including VLM variants)?

tools/visual-pipeline-and-platform-evaluation-tool/vippet/models.py
```diff
@@ -95,18 +95,34 @@ def exists_on_disk(self) -> bool:
...
-            if not exists:
+            if not os.path.isdir(self.model_path_full):
                 logger.debug(
                     f"GenAI model directory not found for '{self.display_name}' at path '{self.model_path_full}'"
                 )
-            return exists
+                return False
+            # Require the main language model file to be present so that an
+            # empty or partially-downloaded directory is not treated as installed.
+            main_model_file = os.path.join(self.model_path_full, "openvino_language_model.xml")
```
kkepka: nitpick: should it use `_GENAI_SENTINEL_FILE` to avoid future changes drift?

KaliszWiktoria: Of course, addressed in: https://github.com/open-edge-platform/edge-ai-libraries/pull/2812/changes/8d79cb4337ec6ec66dda54382e3574d5fa2de7b3

tools/visual-pipeline-and-platform-evaluation-tool/vippet/managers/model_manager.py
KaliszWiktoria: addressed in: https://github.com/open-edge-platform/edge-ai-libraries/commit/8d79cb4337ec6ec66dda54382e3574d5fa2de7b3
(with running commands)

tools/visual-pipeline-and-platform-evaluation-tool/vippet/managers/model_manager.py
KaliszWiktoria: Yes, I think checking for openvino_language_model.xml is a valid _**minimum**_ check that the language-model part of the VLM is installed, in OpenVINO GenAI implementation, VLMPipeline() explicitly looks for:

models_dir / "openvino_language_model.xml"

and then tries to load that file with read_model().

So for the currently supported VLMs, checking whether openvino_language_model.xml exists is in my opinion a good way to check if main language-model is present.

I came across a time when openvino_language_model.xml did not guarantee the correct model installation, but I think it is still better than checking for the main folder and mostly correct.

In the future I might add other checks for other types of models with different files checks.

dlemiech (approved):

KaliszWiktoria (commented):
Future work on changing model installation status indication UI, to use pulling status of model installation from model downloader, will be done for task:
https://jira.devtools.intel.com/browse/ITEP-95663
for 2026.3.0 release

kkepka (approved):

PR #2816, opened 2026-08-13
[ViPPET] benchamark definitions extracted to yaml

dlemiech:
### Description

Application startup will load benchmark definitions from yaml files under: vippet/benchmarks/*.yaml

### Checklist:


tools/visual-pipeline-and-platform-evaluation-tool/vippet/benchmarks/manufacturing-suite.yaml
```diff
@@ -0,0 +1,11 @@
+name: Manufacturing Suite
+description: Manufacturing benchmark suite covering defect detection.
+workloads:
+  - pipeline_id: defect-detection
+    variants:
+      - name: cpu
+        test_cases: [1, 4, 8, 12]
```
kkepka: wording `test_cases` is a bit generic, what do you think about changing name to something more representative, what those "magic numbers" means?

KaliszWiktoria (dismissed):

tools/visual-pipeline-and-platform-evaluation-tool/vippet/db_seed.py
```diff
@@ -2,203 +2,234 @@
...
+    if not benchmark_files:
+        logger.warning(
+            "No benchmark YAML files were found in %s. Expected one suite per file.",
+            benchmarks_dir,
+        )
+        return suite_specs
+
+    for suite_path in benchmark_files:
+        with open(suite_path, "r", encoding="utf-8") as suite_file:
+            loaded = yaml.safe_load(suite_file) or {}
```
kkepka: malformed yaml file might cause app startup crash,
please rather catch errors that might trigger here, log it and proceed gracefully with next files

tools/visual-pipeline-and-platform-evaluation-tool/vippet/db_seed.py
```diff
@@ -2,203 +2,234 @@
...
+                created_at=now,
+                last_run_at=now,
             )
+            session.add(suite)
+            await session.flush()
+        else:
+            suite.name = suite_spec["name"]
+            suite.description = suite_spec["description"]
+
+        for workload_spec in suite_spec["workloads"]:
```
kkepka, line 297: `workloads` and few other keys are assumed to exists,
with malformed/typo/missing structure can cause db seed failure,
please add validation before use, log warnings and skip invalid entries

kkepka (commented):
i like direction of this change overall, please see few other comments
also i assume, that tests coverage will be introduced in next PR, right?

tools/visual-pipeline-and-platform-evaluation-tool/vippet/benchmarks/manufacturing-suite.yaml
dlemiech: testing_streams is ok?

KaliszWiktoria (approved):

tools/visual-pipeline-and-platform-evaluation-tool/vippet/db_seed.py
```diff
@@ -2,203 +2,377 @@
...
-                    slug=suite_slug,
-                    name=suite_spec["name"],
-                    description=suite_spec["description"],
-                    created_at=now,
-                    last_run_at=now,
+        for workload_spec in suite_spec["workloads"]:
+            variant_names = [
+                variant_spec["name"] for variant_spec in workload_spec["variants"]
+            ]
+            variants_value = ",".join(variant_names)
```
kkepka, line 301: what is the intended lifecycle of that (yaml) config, do we expect, that user could edit (e.g. add new variant) after first startup?

dlemiech: right now, the database is loaded only if empty, so adding new benchmarks is not planned after first deploy.

kkepka: ok, thank you for clarification, might be good idea to note that in relevant user documentation, or as comment in config (or readme in configs directory..) will leave it up to you

kkepka: so the seeder should trigger only once, and its gated with [DB_SEED_ON_STARTUP](https://github.com/open-edge-platform/edge-ai-libraries/blob/e4779eb09cbc98d12df9da683768a22159123861/tools/visual-pipeline-and-platform-evaluation-tool/vippet/database.py#L155) var, which is not set anywhere .. ?

kkepka (approved):

PR #2841, opened 2026-08-18
Fix/vss - Addresses HIGH/CRITICAL security scan findings across the VSS sample application and the microservices

bhardwaj-nakul:
### Description

- Addresses HIGH/CRITICAL security scan findings across the VSS sample application and the microservices it depends on. Scans run: Bandit, ClamAV, Trivy (FS/Config/Image), and CodeQL, all with  --ignore-unfixed .

- Result: 97 → 1 HIGH/CRITICAL findings. The single remaining item is  video-ingestion DS-0002, a static-analysis false positive (the  intel/dlstreamer-pipeline-server  base image already runs as uid 1999; Trivy cannot see the base image's USER  directive).

- Dependency refresh (minor/patch only - no major version changes):

    - Regenerated  poetry.lock  for  multimodal-embedding-serving ,  multimodal-dataprep ,  mcp ,  search-ms 
    - Updated  package-lock.json  for  pipeline-manager  and  ui/react ;  pipeline-manager  adds two  overrides  pinning transitive  @opentelemetry/core  and  @opentelemetry/propagator-jaeger  to  ^2.10.0 
    -  cli/go.mod  allows major bumps (no external API consumers)

- OpenVINO stack pinned in  multimodal-embedding-serving  ( openvino <2026.3.0 ,  nncf <3.3.0 ) so lock regeneration can't silently move the inference runtime, which would need validation outside this PR's scope. The pin also covers  multimodal-dataprep , which consumes it via a local path dependency.

- pip removed from production images. Recent pip versions vendor a CycloneDX SBOM declaring  setuptools 70.3.0  and  msgpack 1.1.2 ; neither ships as importable code, but scanners flag them as CVE-2025-47273 and GHSA-6v7p-g79w-8964. Two different fixes were needed:

    -  vector-retriever  installs into the system environment → dropping  pip install --upgrade pip  suffices
    -  mcp  /  multimodal-dataprep  install into a poetry venv that poetry seeds with its own pip → additionally set  POETRY_VIRTUALENVS_OPTIONS_NO_PIP  /  NO_SETUPTOOLS 

Fixes # (issue)

None

### How Has This Been Tested?

Tested in dev env

### Checklist:


sample-applications/video-search-and-summarization/pipeline-manager/package.json
```diff
@@ -99,7 +99,9 @@
     "serialize-javascript": "^7.0.4",
     "ajv": "^8.18.0",
     "file-type": "^21.3.1",
-    "picomatch": "^4.0.4"
+    "picomatch": "^4.0.4",
+    "@opentelemetry/core": "^2.10.0",
```
yogeshmpandey, line 103: is this an override or new deps ?

yogeshmpandey (approved):

sample-applications/video-search-and-summarization/pipeline-manager/package.json
bhardwaj-nakul: override only. identified as high/critical vuln.

krish918 (approved):

PR #2921, opened 2026-08-21
fix: RTmotion - motion blending produces incorrect trajectory transitions

### LINKED ISSUE #2894: [BUG] RTmotion - motion blending produces incorrect trajectory transitions
### Bug Description

The RTmotion blending mode does not provide the expected continuous motion behavior. When executing consecutive motion segments using `BlendingLow`, the resulting trajectory behavior does not achieve the expected throughput and smooth transition characteristics.

1. Incomplete (move never finishes)
    - **What happens**: move1 never reports `Done`, so move2 (queued behind it) never activates — the axis just cruises forever at the blend velocity.
    - **Expected behavior**: move1 should report `Done` as soon as it reaches (or passes) its blended end position/velocity, so move2 can take over immediately, per PLCopen Part 1&2 §5.4.

2. Exception / FbError
    - **What happens**: the axis never starts moving at all — the function block reports an error (`isError()==mcTRUE`) within milliseconds of queuing.
    - **Expected behavior**: move1 should ramp up past its own nominal velocity toward the higher blended velocity, then hand off to move2 smoothly (or, if that is judged infeasible under move1's configured limits, fail only after motion is attempted — not before it starts).

3. Velocity glitch at handoff
    - **What happens**: a one-cycle velocity spike/dip right at the blend boundary, even once defects issue 1 and 2 above are fixed.
    - **Expected behavior**: velocity should be continuous across the blend boundary — the whole point of blending is that the axis never stops or jumps between the two moves.

### System Setup

- OS: Ubuntu 22.04

### Reproducible Steps

1. Configure RTMotion on the system described above.
2. Create a trajectory consisting of multiple consecutive motion segments requiring blending.
3. Execute the trajectory using `BlendingLow` mode.
4. Measure transition behavior and overall throughput.
5. Repeat execution multiple times.

### Root Cause Analysis

1. Incomplete (move never finishes)
    - **Root cause**: `ExecutionNode::checkMissionDone()` (`library/fb/common/src/execution_node.cpp`) requires position AND velocity to be within a fixed `+/-0.0001` tolerance band in the same cycle. Whenever the per-cycle displacement (axis velocity × cycle time) exceeds that band width — e.g. at a high blended velocity, at any cycle rate — the axis can cross the band entirely within one cycle, so `checkMissionDone()` never samples inside it.

2. Exception / FbError
    - **Root cause**: `MotionKernel::addFBToQueue()` (`library/fb/common/src/motion_kernel.cpp`) sets move1's blended end-velocity to `vel2` (`mcBlendingNext`) or `max(vel1, vel2)` (`mcBlendingHigh`) without raising move1's own configured velocity limit (used as the ruckig planner's `max_velocity`). When the blend velocity exceeds that limit, the planner rejects the segment outright.

3. Velocity glitch at handoff
    - **Root cause**: `MotionKernel::addFBToQueue()` seeds move2's start position/velocity from move1's idealized planned end target (`fb_prev->getEndPos()`/`getEndVel()`), not the axis's actual position at the crossing cycle. Since fix issue 1 intentionally lets the axis overshoot slightly before `Done` fires, move2 then starts from a position slightly behind where the axis actually is, and the virtual servo's finite-difference velocity derivation turns that tiny position snap into a visible one-cycle velocity spike/dip. The size of that gap grows with cycle time (a larger `dt` means more overshoot accumulates before the crossing check in fix issue 1 fires), so this glitch is more pronounced at lower cycle rates, not specific to any one rate.

mazlanmu:
### Description

This pull request addresses a blending bug in the motion control library, improving the reliability and accuracy of move completion detection and blending handoff between motion segments. The main changes ensure that the system properly detects when a target position is crossed within a single control cycle (even if the done-tolerance band is skipped), and that blending handoffs use actual axis feedback to avoid velocity glitches. Additionally, blending logic now updates velocity limits to accommodate higher blend velocities.

**Blending and Move Completion Improvements:**

* `execution_node.hpp`, `execution_node.cpp`: Added `prev_pos_err_` and `has_prev_pos_err_` to track previous position error, allowing `checkMissionDone()` to detect when the target is crossed within a single cycle—even if the done-tolerance band is skipped. This fixes missed mission completion events during blending.
* `execution_node.cpp`: Reset `has_prev_pos_err_` when a new trajectory segment starts to avoid false crossing detection from stale data.

**Blending Handoff and Velocity Handling:**

* `motion_kernel.cpp`: During blending handoff, re-seed the next node's start position and velocity from actual axis feedback (not just the previous FB's idealized target) to prevent handoff velocity glitches, especially at lower cycle rates.
* `motion_kernel.cpp`: In blending modes, if the blend velocity exceeds the previous node's velocity limit, raise the limit to match the blend velocity, ensuring the planner does not reject the blend and the handoff is smooth.

**Fixes # (issue)**
- closes #2894 

n/a

### How Has This Been Tested?

- Ideally, the current RTmotion native test should cover blending mode as well. Assuming there is native test, it should built directly against RTmotion's own -DTEST=ON unit tests in edge-ai-libraries. Both cases are fail against the unfixed library and pass with the fix applied.
- Verified via BlendingBufferModeTest.*_Incomplete/*_Exception in our custom test suite, with per-cycle velocity/position CSVs and plots confirming continuous, non-zero velocity across the blend boundary and no handoff glitch. 

### Checklist:


RoboticsYY:
Hi @mazlanmu, Thanks for your fix. Could you please resubmit the PR to https://github.com/intel-innersource/frameworks.industrial.edge-controls.components.plcopen-motion, that's our development branch, we use this location for release. Could you please also provide an example test case to reproduce the issue? Then I can look into it.

mohitmeh12 (approved):

jouillet (approved):

PR #2927, opened 2026-08-21
Fix/vss - Fix vLLM model guidance, search UI defects, and a vector-retriever concurrency hang

bhardwaj-nakul:
### Description

**1. vLLM model guidance pointed at an unsupported model.** A broad-stroke update had set
`VLM_MODEL_NAME` to Qwen3-VL everywhere, but vLLM does not serve it, so `ENABLE_VLLM=true`
deployments failed to come up. Reverted to Qwen2.5-VL in vLLM-specific docs, skills, and
env templates only; OVMS paths keep Qwen3-VL.

**2. Group by Tag dropped search results.** Grouped view deduplicated by `videoId`, so 20
hits across 3 videos rendered as 3 cards. Now groups per search result, keeping every hit.

**3. New search queries were invisible until results arrived.** The sidebar filtered to
entries with results, so a slow query looked like nothing happened. Removed the filter and
split cold-start (skeleton) from watched-query refresh (results stay mounted), so the
results area no longer flickers empty on every refresh.

**4. Concurrent searches could hang the vector-retriever.** All requests share one cached
`VDMS_Client` whose socket send/recv is not atomic, so concurrent queries interleaved
frames, one thread stole another's length prefix, and the rest blocked in `recv()`
forever, requiring a container restart. Serialized the two VDMS entry points with a
`threading.RLock`, applied only to the VDMS backend so pooled Milvus and PGVector stay
parallel. Measured 1/5 queries surviving before, 15/15 after, with no throughput cost.

**5. Search modal ignored Enter and accepted empty queries.** Enter inserted a newline
instead of searching, and an empty query closed the modal and ran a meaningless search.
Enter now submits (Shift+Enter still newlines, IME respected) and empty input shows an
inline error without closing.

**6. Time filter was unclear and could be read as video duration.** The label was the
fragment "Added within the last", styled identically to its summary line, and actually
filters on ingestion time rather than video length. Restructured as a "Date added"
fieldset with demoted helper text, reworded the tooltip, disabled the unit dropdown at 0,
fixed the tooltip being clipped by the modal, and converted UI strings to ASCII after an
em dash silently desynced a translation from its fallback.

Fixes # (issue)

None

### How Has This Been Tested?

Tested in dev env

### Checklist:


yogeshmpandey (approved):

krish918 (approved):

yogeshmpandey (dismissed):

madhuri-rai07 (dismissed):

microservices/vector-retriever/vector-retriever/src/retriever/service.py
```diff
@@ -30,6 +32,33 @@
 logger = get_logger()
 TIME_FILTER_METADATA_FIELD = "created_at"
 
+# The VDMS backend caches a single process-wide ``VDMS_Client`` (see
```
bharagha: Just a check: Shouldnt this be more moved to VDMS to manage?

sample-applications/video-search-and-summarization/.github/skills/vss-deploy-helm/SKILL.md
```diff
@@ -270,7 +270,7 @@ service, and `global.devices.multimodalDataprep.detection` for DataPrep object
 detection. These settings are independent; every GPU/NPU setting requires its
 own resource `key`.
 
-vLLM tuning keys from the actual `vllm` subchart:
+vLLM tuning keys from the actual `vllm` subchart (set `global.vlmName: "Qwen/Qwen2.5-VL-3B-Instruct"` for vLLM):
```
bharagha: Need to check if specifying the model means recommending the model.
When a model is not supported, shouldn't it be more managed with error handling?

BTW, check the vLLM image tag if it is using the right latest version

bharagha (dismissed):

sample-applications/video-search-and-summarization/.github/skills/vss-deploy-helm/SKILL.md
bhardwaj-nakul: Removed the model recommendation from skill file.

vllm version update will be done in upcoming weeks.

microservices/vector-retriever/vector-retriever/src/retriever/service.py
bhardwaj-nakul: Moved this func to vdms specific directory.

yogeshmpandey (approved):

llin60 (approved):

PR #3007, opened 2026-08-27
[ViPPET] Fixes issue with buffering, where nginx gated metadata updated on their way to browser/UI (ITEP-93483)

kkepka:
### Description

- Pad metadata SSE so low-rate streams are not withheld 
- Answer 204 on the metadata stream of a finished job 
- Disable proxy buffering on the nginx /api/ location

Fixes issue with buffering, where nginx gated metadata updated on their way to browser/UI.

None

### How Has This Been Tested?

Manually on at least two different deployments by two engineers.

### Checklist:


KaliszWiktoria (approved):

tools/visual-pipeline-and-platform-evaluation-tool/ui/nginx.conf
```diff
@@ -21,6 +21,13 @@ server {
         proxy_set_header Access-Control-Allow-Methods "GET, POST, PUT, DELETE, OPTIONS";
         proxy_set_header Access-Control-Allow-Headers "Content-Type, Authorization";
 
+        # The API serves SSE (job metadata streams) on this prefix, so responses
```
dlemiech, line 24: this block is for REST api, I think there should not be any SEE activities here

we should review this workflow where we attach SEE endpoints to api endpoints

dlemiech (approved):

tools/visual-pipeline-and-platform-evaluation-tool/ui/nginx.conf
kkepka: good idea, we could isolate streams to its own uri/configuration, lets follow-up on this

PR #3095, opened 2026-09-02
[VSS] Fix for mme/dataprep

bhardwaj-nakul:
#### Problem

Video embedding requests failed with `HTTP 500` / `[Errno 24] Too many open files` at large frame batch sizes. Three related defects:

1. **fd exhaustion** — `SharedMemoryPool` kept an open handle per block (2 fds each) for the pool's lifetime, so a 512-block pool exceeded the 1024 fd limit before decoding started. `multimodal-dataprep` forks the same pool and builds *two*   (frames + 2× crops, ~3000 fds), so it was worse there.
2. **/dev/shm leak** — teardown wrapped the whole loop in one `try`, so a single already-unlinked block aborted cleanup and leaked the remaining segments.
3. **GPU defaults never applied** — `setup.sh` assigned batch-size defaults *before* testing `EMBEDDING_DEVICE`, leaving the GPU branch's `${VAR:-...}` fallbacks dead. GPU deployments silently ran at CPU batch sizes.

#### Fix

- Close the handle right after allocation — the segment lives until `unlink()` and every stage attaches by name. **512 blocks: 1024 fds → 1.**
- Unlink each block independently.
- Resolve batch-size defaults per device before assignment. GPU now uses `INFER_BATCH_SIZE=32` / `VIDEO_FRAME_BATCH_SIZE=256` (fastest measured for CLIP ViT-B/32); explicit user values still win. Device matching now accepts indexed devices such as `GPU.0`, which the docs already documented but which previously fell through to the CPU path.
- VSS never set the frame batch size at all; its GPU path now selects 256.
- Raise `nofile` on the embedding-serving containers, matching the limit  `multimodal-dataprep` already had.

Fixes # (issue)

None

### How Has This Been Tested?

Tested in dev env.

### Checklist:


yogeshmpandey (approved):

microservices/multimodal-embedding-serving/src/utils/decoder.py
```diff
@@ -75,8 +75,12 @@ def __init__(self, max_blocks, block_size):
         self.blocks = []
         self.in_use = set()
 
+        # An open handle costs 2 fds for the pool's lifetime, so a large pool
+        # would exhaust the fd limit. The segment lives until unlink() and is
+        # attached by name on demand, so release the descriptor here.
         for _ in range(max_blocks):
             shm = shared_memory.SharedMemory(create=True, size=block_size)
+            shm.close()
```
bharagha, line 83: Shouldnt close be after append and put?

microservices/visual-data-preparation-for-retrieval/multimodal-dataprep/src/core/embedding/decoder.py
```diff
@@ -99,8 +99,12 @@ def __init__(self, max_blocks, block_size):
         self.in_use = set()
         self._lock = threading.Lock()
 
+        # An open handle costs 2 fds for the pool's lifetime. This pipeline
+        # builds two pools (frames plus twice as many crops), which would pin
+        # ~3000 fds. The segment lives until unlink() and is attached by name.
         for _ in range(max_blocks):
             shm = shared_memory.SharedMemory(create=True, size=block_size)
+            shm.close()
```
bharagha, line 107: Same comment

microservices/multimodal-embedding-serving/src/utils/decoder.py
bhardwaj-nakul: shm.close() only releases the file descriptor - it doesn't destroy the segment, and .name remains valid (CPython's  close() never touches  _name ), so publishing the name afterwards is safe and keeps every tracked block descriptor-free from the moment it's created.

microservices/visual-data-preparation-for-retrieval/multimodal-dataprep/src/core/embedding/decoder.py
bhardwaj-nakul: shm.close() only releases the file descriptor - it doesn't destroy the segment, and .name remains valid (CPython's  close() never touches  _name ), so publishing the name afterwards is safe and keeps every tracked block descriptor-free from the moment it's created.

bharagha (approved):

